PR #3, opened 2016-02-24
Rest processor

purple52:
Add REST processor for generated JAX-RS classes


rest-adapter/src/main/java/uk/gov/justice/services/adapter/rest/envelope/RandomUUIDGenerator.java
```diff
@@ -0,0 +1,17 @@
...
+ * Factory for creating new UUIDs.
+ */
+public class RandomUUIDGenerator {
+
+    /**
+     * Generate a random UUID.
+     * @return a UUID
+     */
+    public UUID generate() {
+        return UUID.randomUUID();
```
kolektiv: This seems like a slightly elaborate wrapper...? I presume that this is to make something testable via mocking? If so, could we change the responsibilities somewhere to accept a UUID rather than generating one, to simplify?

kolektiv: (If you read the previous comment with an Australian accent, all of the question marks make sense?)

PR #10, opened 2016-03-02
Refactor core handler test

mapingo:


core/src/test/java/uk/gov/justice/services/core/handler/HandlerInstanceAndMethodTest.java
```diff
@@ -29,44 +27,48 @@
     private CommandHandler commandHandler;
 
     @Test
-    public void shouldExecuteHandlerMethod() throws NoSuchMethodException, IllegalAccessException, InvocationTargetException, IOException {
+    public void shouldExecuteHandlerMethod() throws Exception {
         Envelope envelope = testEnvelope("envelope.json");
```
matt-rich: Could make this value a constant as it is reused.

core/src/test/java/uk/gov/justice/services/core/handler/HandlerInstanceAndMethodTest.java
```diff
@@ -29,44 +27,48 @@
...
+
+    private HandlerInstanceAndMethod handlerInstanceWithMethod() {
+        return new HandlerInstanceAndMethod(commandHandler, methods().get(0));
+    }
+
+    private List<Method> methods() {
+        return HandlerUtil.findHandlerMethods(CommandHandler.class, Handles.class);
     }
 
     private Envelope testEnvelope(String fileName) throws IOException {
```
matt-rich: I think we said that we would always use braces rather than flattening to one line as SonarQube reports on this.

Is there a better name than handler1 for the method? - make it clearer.

core/src/test/java/uk/gov/justice/services/core/handler/HandlerInstanceAndMethodTest.java
mapingo: Moved to @Before and set field.

core/src/test/java/uk/gov/justice/services/core/handler/HandlerInstanceAndMethodTest.java
mapingo: Cleaned up CommandHandler class

PR #13, opened 2016-03-03
Add metrics library to provide health check endpoint

purple52:


metrics/src/main/java/metrics/MetricsServletContextListener.java
```diff
@@ -0,0 +1,21 @@
...
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.servlets.MetricsServlet;
+
+import javax.servlet.annotation.WebListener;
+
+/**
+ * Annotated context listener for wiring up the metrics servlet.
+ */
+@WebListener
+public class MetricsServletContextListener extends MetricsServlet.ContextListener {
```
matt-rich: Should this be an abstract class? Or is it just part of how the drop wizard pulls it in?

metrics/src/main/java/metrics/HealthCheckServletContextListener.java
```diff
@@ -0,0 +1,21 @@
...
+import com.codahale.metrics.health.HealthCheckRegistry;
+import com.codahale.metrics.servlets.HealthCheckServlet;
+
+import javax.servlet.annotation.WebListener;
+
+/**
+ * Annotated context listener for wiring up the health check servlet.
+ */
+@WebListener
+public class HealthCheckServletContextListener extends HealthCheckServlet.ContextListener {
```
matt-rich: static import for HealthCheckServlet.ContextListerner

matt-rich: Should this be an abstract class with static getter? Or is this just how it has to be to work with the framework?

metrics/src/main/java/metrics/MetricsServletContextListener.java
purple52: This needs to implement the abstract non-static getter required by the abstract ContextListener Dropwizard class.

metrics/src/main/java/metrics/HealthCheckServletContextListener.java
purple52: This needs to implement the abstract non-static getter required by the abstract ContextListener Dropwizard class.

PR #20, opened 2016-03-16
Crc 1531 - event listener generation, validation and integration tests

jaceko:
Renaming JMS->messaging still pending but I would like to integrate the changes into master first.
There is also a potential bug in the plugin around processing multiple RAMLs that needs to be investigated.


adapters/jms-adapter-generator/pom.xml
```diff
@@ -7,6 +7,31 @@
         <version>0.1.0-SNAPSHOT</version>
     </parent>
     <artifactId>jms-adapter-generator</artifactId>
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
purple52: Build source encoding is set in the parent POM, so shouldn't be needed here.

adapters/jms-adapter-generator/pom.xml
```diff
@@ -7,6 +7,31 @@
...
+        <version.org.jboss.arquillian>1.1.11.Final</version.org.jboss.arquillian>
+        <version.org.wildfly>10.0.0.Final</version.org.wildfly>
```
purple52: We should set these versions in the parent POM.

adapters/jms-adapter-generator/pom.xml
```diff
@@ -7,6 +7,31 @@
...
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
+            <dependency>
+                <groupId>org.wildfly.bom</groupId>
+                <artifactId>wildfly-javaee7</artifactId>
+                <version>${version.org.wildfly}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
```
purple52: These imports could move to the parent POM

adapters/jms-adapter-generator/pom.xml
```diff
@@ -56,6 +81,167 @@
             <artifactId>commons-testing</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.arquillian</groupId>
+            <artifactId>wildfly-arquillian-container-managed</artifactId>
+            <version>1.0.2.Final</version>
```
purple52: Version should be set in dependency management, preferably in parent POM

adapters/jms-adapter-generator/pom.xml
```diff
@@ -56,6 +81,167 @@
...
+                                        </goals>
+                                    </pluginExecutionFilter>
+                                    <action>
+                                        <ignore></ignore>
+                                    </action>
+                                </pluginExecution>
+                            </pluginExecutions>
+                        </lifecycleMappingMetadata>
+                    </configuration>
+                </plugin>
```
purple52: Can we remove Eclipse-specific configuration?

adapters/jms-adapter-generator/pom.xml
```diff
@@ -56,6 +81,167 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>enforce-rules</id>
+                        <phase>none</phase>
+                    </execution>
+                </executions>
+            </plugin>
```
purple52: Don't need to switch of the enforcer if you use the release version of the plugin.

jaceko: we use snapshot because the master branch does not seem to be migrated to the released version yet

adapters/jms-adapter-generator/pom.xml
jaceko: Not sure if It's causing any harm and it helps to develop on Eclipse. I remember the exact bit being present in the old parent pom.
We could move it to the new parent pom.

adapters/jms-adapter-generator/pom.xml
purple52: Then update to use the release version... Talk to Justin; he has pom updates for maven-common that could be pushed if you need them

PR #46, opened 2016-04-26
Add aggregate service and helper functions

purple52:


core/src/main/java/uk/gov/justice/services/core/aggregate/AggregateService.java
```diff
@@ -0,0 +1,62 @@
...
+     * Recreate an aggregate of the specified type by replaying the events from an event stream.
+     * @param stream the event stream to replay
+     * @param clazz the type of aggregate to recreate
+     * @param <T> the type of aggregate being recreated
+     * @return the recreated aggregate
+     */
+    public <T extends Aggregate> T get(final EventStream stream, final Class<T> clazz) {
+
+        try {
+            T aggregate = clazz.newInstance();
```
mapingo: "T aggregate" can be final

domain/src/main/java/uk/gov/justice/domain/aggregate/condition/Precondition.java
```diff
@@ -0,0 +1,53 @@
+package uk.gov.justice.domain.aggregate.condition;
+
+import java.util.function.Supplier;
+
+/**
+ * Representation of a precondition that can be asserted, and throw an exception is it isn't met.
+ */
```
mapingo: "...exception is it isn't met." should read "...exception if it isn't met"

PR #49, opened 2016-05-04
Convert Rest Adapter Generator to use Poet

mapingo:


adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/Generators.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Comparator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.raml.model.MimeType;
+import org.raml.model.Raml;
+
+final class Generators {
+    private static final Pattern PILLAR_AND_TIER_PATTERN = Pattern
+            .compile("(command/api|command/controller|command/handler|query/api|query/controller|query/view)");
```
matt-rich: Events should be here also (Listener and Processor)

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/Generators.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private Generators() {
+    }
+
+    static Component componentFromBaseUriIn(final Raml raml) {
```
matt-rich: Method comment

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/Generators.java
```diff
@@ -0,0 +1,41 @@
...
+import uk.gov.justice.services.core.annotation.Component;
+
+import java.util.Comparator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.raml.model.MimeType;
+import org.raml.model.Raml;
+
+final class Generators {
```
matt-rich: Should this be public or is this intentional?

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsApplicationCodeGenerator.java
```diff
@@ -34,67 +34,91 @@
...
-    private void addAnnotations(final Raml raml, final JDefinedClass application) {
-        application.annotate(ApplicationPath.class)
-                .param(DEFAULT_ANNOTATION_PARAMETER, defaultIfBlank(baseUriPathWithoutContext(raml), "/"));
+    /**
+     * Generate the implementation of {@link Application} and set the {@link ApplicationPath}.
+     *
+     * @param raml the RAML document being generated from
+     * @return the {@link TypeSpec.Builder} that defines the class
+     */
+    private TypeSpec.Builder classSpecFor(final Raml raml) {
```
matt-rich: ClassSpecFrom is a more suitable name here?

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsApplicationCodeGenerator.java
```diff
@@ -34,67 +34,91 @@
...
+        return classBuilder(applicationNameOf(raml))
```
matt-rich: applicationNameFrom might be a nicer method name.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsApplicationCodeGenerator.java
```diff
@@ -34,67 +34,91 @@
...
-        JType classSetType = codeModel.ref(Set.class).narrow(wildcardClassType);
-
-        JMethod getClassesMethod = application.method(PUBLIC, classSetType, "getClasses");
+    /**
+     * Generate code to add each resource implementation class to the classes hash set.
+     *
+     * @param implementationNames the collection of implementation class names
+     * @return the {@link CodeBlock} that represents the generated statements
+     */
+    private CodeBlock codeStatementForEach(final Collection<String> implementationNames) {
```
matt-rich: More descriptive method name? - What is the code statement being added (not clear from calling method)

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsImplementationGenerator.java
```diff
@@ -0,0 +1,348 @@
...
+     * Process the body or bodies for each action.
+     *
+     * @param action the action to process
+     * @return the list of {@link MethodSpec} that represents each method for the action
+     */
+    private List<MethodSpec> forEach(final Action action) {
+        final boolean actionHasNoBody = !action.hasBody();
+
+        if (actionHasNoBody) {
+            return processNoActionBody(action);
```
matt-rich: move Collections.SingletonList up to here as it is this method that requires it to be a list, the method should just return the single value.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsImplementationGenerator.java
```diff
@@ -0,0 +1,348 @@
...
+    private MethodSpec.Builder generateResourceMethod(final Action action,
+                                                      final String resourceMethodName) {
+        final ActionType actionType = action.getType();
+        final Map<String, QueryParameter> queryParams = action.getQueryParameters();
+        final Map<String, UriParameter> pathParams = action.getResource().getUriParameters();
+        final CodeBlock methodBody;
+
+        if (actionType == GET) {
+            methodBody = methodBody(pathParams, () -> methodBodyForGet(queryParams));
+        } else {
```
matt-rich: This should be updated to check if it is post and if anything else  throw an exception, see method above for how this is done.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
```diff
@@ -0,0 +1,262 @@
...
+     * Process the body or bodies for each action.
+     *
+     * @param action the action to forEach
+     * @return the list of {@link MethodSpec} that represents each method for the action
+     */
+    private List<MethodSpec> forEach(final Action action) {
+        final Collection<MimeType> responseMimeTypes = responseMimeTypesOf(action);
+        final boolean actionHasNoBody = !action.hasBody();
+
+        if (actionHasNoBody) {
```
matt-rich: No reason to pull this into a local variable.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    /**
+     * Process an action with no body.
+     *
+     * @param action the action to process
+     * @return the list of {@link MethodSpec} that represents each method for the action
+     */
+    private List<MethodSpec> processNoActionBody(final Action action,
```
matt-rich: return single not list?

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
```diff
@@ -0,0 +1,262 @@
...
+                                                      final String resourceMethodName,
+                                                      final Collection<MimeType> responseMimeTypes) {
+
+        final ActionType actionType = action.getType();
+        final Map<String, QueryParameter> queryParams = action.getQueryParameters();
+        final Map<String, UriParameter> pathParams = action.getResource().getUriParameters();
+        final AnnotationSpec actionTypeAnnotation;
+
+        if (actionType == GET) {
+            actionTypeAnnotation = AnnotationSpec.builder(javax.ws.rs.GET.class).build();
```
matt-rich: we should static import these GET, POST

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /**
+     * Generates the Produces annotation for all the response mime types.
+     *
+     * @param responseMimeTypes generate annotations for each responseMimeType
+     * @return the list of {@link AnnotationSpec} that represent the produces annotations
+     */
+    private List<AnnotationSpec> annotationsForProduces(final Collection<MimeType> responseMimeTypes) {
+        final List<AnnotationSpec> specs = new ArrayList<>();
```
matt-rich: Perhaps we shouldn't make this list final as we are modifying it below (it is the entire intention of this method).

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/Generators.java
mapingo: This was intentional, I wanted to keep the utils classes package local unless otherwise needed.  We do need to be consistent though so I will make sure the other utils classes are set the same way.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/Generators.java
mapingo: Events don't make any sense for the Rest adapter generator.  Do we want to move this into the framework as it is specific to the RAML to adapter generation.  Worth some further discussion.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/Generators.java
mapingo: Done

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsApplicationCodeGenerator.java
mapingo: Yep, better name. Done.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsApplicationCodeGenerator.java
mapingo: Done

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/RestAdapterGenerator.java
```diff
@@ -63,4 +68,53 @@ private void validate(final GeneratorConfig configuration) {
...
+     * @param typeSpecs     the list of typeSpecs to write to file
+     * @param configuration the configuration that provides the base package name
+     * @return the list of class names written to file
+     * @throws IllegalStateException if an IOException is thrown while writing to file
+     */
+    private List<String> writeToBasePackage(final List<TypeSpec> typeSpecs,
+                                            final GeneratorConfig configuration,
+                                            final String packageName) {
+
+        final List<String> implementationNames = new ArrayList<>();
```
matt-rich: Shouldn't be final, as reason above.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsApplicationCodeGenerator.java
mapingo: Difficult to come up with a descriptive name here, but done.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsImplementationGenerator.java
mapingo: Done.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsImplementationGenerator.java
mapingo: The problem here is that the action type is already checked before this method is run.  So, if I add the check here it becomes untestable code.  I have decided to make it clearer what is occurring and I have added a method to validate the action type prior to processing the action.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
mapingo: Done

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
mapingo: Done

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
mapingo: Unable to static import as the ActionType.GET is already imported

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
mapingo: I think it should be fine as final, we are just protecting the reference is not overwritten in this case

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/RestAdapterGenerator.java
mapingo: Same as above, no harm in setting to final, even if the internal representation of the object itself is updated

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsInterfaceGenerator.java
```diff
@@ -73,8 +72,10 @@
     private TypeSpec createInterfaceFor(final Resource resource) {
         final TypeSpec.Builder interfaceSpecBuilder = interfaceSpecFor(resource);
 
-        resource.getActions().values().forEach(action ->
-                interfaceSpecBuilder.addMethods(forEach(action)));
+        resource.getActions().values().forEach(action -> {
```
matt-rich: Could we use a filter to remove those that are not valid rather than encapsulating the check within?

mapingo: Quite right...I suppose it depends whether we want to throw an IllegalStateException and stop processing the RAML or just ignore other types.  I will go for ignoring the other types and filter.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsImplementationGenerator.java
```diff
@@ -94,6 +93,7 @@ private TypeSpec createClassFor(final Resource resource, final Component compone
         final TypeSpec.Builder classSpecBuilder = classSpecFor(resource, component);
 
         resource.getActions().values().forEach(action -> {
+            validateActionTypeIsGetOrPost(action.getType());
```
matt-rich: validateActionTypeIsSupported is a better name for this method in-case we update it later.

adapters/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/JaxRsImplementationGenerator.java
```diff
@@ -191,10 +188,8 @@ private FieldSpec dispatcherFieldFor(final Action action) {
 
         if (actionType == GET) {
```
matt-rich: We think that this code was okay before with the if/else if/else throw exception as if we added a new supported type (e.g. PUT) then it could potentially fail silently.

PR #65, opened 2016-05-18
Add support for remote service invocation

ZeeshanGhalib:


clients/rest-client-core/src/main/java/uk/gov/justice/services/clients/core/RestClientProcessor.java
```diff
@@ -23,6 +28,7 @@
 public class RestClientProcessor {
 
     private static final String MEDIA_TYPE_PATTERN = "application/vnd.%s+json";
+    private static final String CPPID = "CPPID";
```
matt-rich: Is it worth reusing the header constant that defines this?

adapters/rest-adapter-core/src/main/java/uk/gov/justice/services/adapter/rest/RestProcessor.java
```diff
@@ -38,9 +39,14 @@
 
     private final Function<JsonEnvelope, String> responseBodyGenerator;
 
-    RestProcessor(RestEnvelopeBuilderFactory envelopeBuilderFactory, Function<JsonEnvelope, String> responseBodyGenerator) {
+    private final boolean sendMetadataIdInHeader;
+
+    RestProcessor(final RestEnvelopeBuilderFactory envelopeBuilderFactory,
+                  final Function<JsonEnvelope, String> responseBodyGenerator,
+                  final boolean sendMetadataIdInHeader) {
```
purple52: Maybe we should pass in a function to add headers, rather than a boolean - to avoid a proliferation of flag arguments

clients/rest-client-core/src/main/java/uk/gov/justice/services/clients/core/RestClientProcessor.java
ZeeshanGhalib: Yes, I wanted to do that but it will require dependency on another module.  We should probably move that HeaderConstants class to rest-client-core or some kind of common module.  Requires some refactoring of that area.

adapters/rest-adapter-core/src/main/java/uk/gov/justice/services/adapter/rest/RestProcessor.java
ZeeshanGhalib: As discussed, we will leave it as is for now.

PR #66, opened 2016-05-23
Check for missing CPPID in request headers

justin-r-oliver:
The current code throws a NullPointerException if the response does not contain a CPPID header. This can occur if the service has been generated outside the framework e.g. a mock service used for integration testing. This change checks for a missing CPPID header and throws an IllegalStateException with a meaningful exception message.

### DIFF
--- a/clients/rest-client-core/src/test/java/uk/gov/justice/services/clients/core/RestClientProcessorIT.java
+++ b/clients/rest-client-core/src/test/java/uk/gov/justice/services/clients/core/RestClientProcessorIT.java
@@ -173,6 +173,28 @@ public class RestClientProcessorIT {
         restClientProcessor.request(endpointDefinition, requestEnvelopeParamAParamB());
     }
 
+    @Test(expected = RuntimeException.class)
+    public void shouldThrowExceptionOnCPPIDMissing() throws Exception {
+
+        final String path = "/my/resource";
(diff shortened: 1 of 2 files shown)


clients/rest-client-core/src/test/java/uk/gov/justice/services/clients/core/RestClientProcessorIT.java
```diff
@@ -173,6 +173,28 @@ public void shouldThrowExceptionOnQueryParamMissing() throws Exception {
...
+                .willReturn(aResponse()
+                        .withStatus(200)
+                        .withHeader("Content-Type", mimetype)
+                        .withBody(envelopeWithoutMetadataAsJson)));
+
+        Set<QueryParam> queryParams = ImmutableSet.of(new QueryParam("paramA", true), new QueryParam("paramB", false), new QueryParam("paramC", true));
+
+        EndpointDefinition endpointDefinition = new EndpointDefinition(BASE_URI, path, emptySet(), queryParams);
+
+        validateResponse(restClientProcessor.request(endpointDefinition, requestEnvelopeParamAParamC()), envelopeWithMetadataAsJson);
```
ZeeshanGhalib: I don't think we need to call validateResponse here.  restClientProcessor.request will throw the exception.

purple52:
Is it an illegal state if the header is missing, though?

PR #70, opened 2016-05-24
Check for missing CPPID in request headers

justin-r-oliver:
The current code throws a NullPointerException if the response does not contain a CPPID header. This can occur if the service has been generated outside the framework e.g. a mock service used for integration testing. This change checks for a missing CPPID header and throws an IllegalStateException with a meaningful exception message.

### DIFF
--- a/clients/rest-client-core/src/main/java/uk/gov/justice/services/clients/core/RestClientProcessor.java
+++ b/clients/rest-client-core/src/main/java/uk/gov/justice/services/clients/core/RestClientProcessor.java
@@ -105,6 +105,10 @@ public class RestClientProcessor {
             return responseAsJsonObject;
         }
 
+        if (cppId == null) {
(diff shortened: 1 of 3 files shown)


clients/rest-client-core/src/main/java/uk/gov/justice/services/clients/core/RestClientProcessor.java
```diff
@@ -91,6 +91,10 @@ private JsonObject addMetadataIfMissing(final JsonObject responseAsJsonObject, f
             return responseAsJsonObject;
         }
 
+        if (cppId == null) {
+            throw new RuntimeException(format("%s is required in request header", CPPID));
```
ZeeshanGhalib: I discussed this with Andrew and he said that we should return 400 without any body.

PR #76, opened 2016-05-27
Support field level ServiceComponent annotation

ZeeshanGhalib:


core/src/main/java/uk/gov/justice/services/core/annotation/Component.java
```diff
@@ -73,7 +78,17 @@ public static Component componentFrom(final Class<?> clazz) {
      * @return the component from the provided injection point
      */
     public static Component componentFrom(final InjectionPoint injectionPoint) {
-        return componentFrom(injectionPoint.getMember().getDeclaringClass());
+        return fieldLevelComponent(injectionPoint).orElseGet(() -> componentFrom(injectionPoint.getMember().getDeclaringClass()));
```
mapingo: Might be more readable on multiple lines

PR #80, opened 2016-05-30
Code change for review comment

hasankshaik:


common/src/main/java/uk/gov/justice/services/common/converter/ListToJsonArrayConverter.java
```diff
@@ -0,0 +1,44 @@
...
+	public JsonArray convert(final List<T> sourceList) {
+		JsonArrayBuilder jsonArrayBuilder = createArrayBuilder();
+		if (sourceList == null) {
+			throw new ConverterException(format("Failed to convert %s to JsonArray", sourceList));
+		}
+		sourceList.forEach((object) -> {
+			try {
+				jsonArrayBuilder.add(stringToJsonObjectConverter.convert(mapper.writeValueAsString(object)));
+			} catch (IOException e) {
+				throw new IllegalArgumentException(format("Error while converting %s JsonArray", object), e);
```
ZeeshanGhalib: Better to throw ConverterException here.  
Also, the message can be improved I think.

common/src/test/java/uk/gov/justice/services/common/converter/ListToJsonArrayConverterTest.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	@Test
+	public void shouldThrowIllegalArgumentException() throws JsonProcessingException {
+		ListToJsonArrayConverter<Pojo> listToJsonArraysConverter = new ListToJsonArrayConverter<Pojo>();
+		listToJsonArraysConverter.mapper = mapperMock;
+		listToJsonArraysConverter.stringToJsonObjectConverter = new StringToJsonObjectConverter();
+
+		Pojo pojoOne = new Pojo(null, null);
+		doThrow(JsonProcessingException.class).when(mapperMock).writeValueAsString(pojoOne);
```
ZeeshanGhalib: Shouldn't we throw IOException here, because that's the top level exception thrown by writeValueAsString and that's what we are catching in our code?

PR #94, opened 2016-06-14
Action mapping update

mapingo:


adapters/adapters-commons/src/main/java/uk/gov/justice/raml/common/mapper/ActionMapping.java
```diff
@@ -77,6 +91,21 @@
         return emptyList();
     }
 
+    private static String extractMappingFrom(final String mappingString) {
+        return removeUpToIncludingMappingBoundary(removeAfterIncludingMappingBoundary(mappingString));
+    }
+
+    private static String removeUpToIncludingMappingBoundary(final String source) {
```
matt-rich: Would trimPreMappingDescription() and trimPostMappingDescription() be clearer names? - or is there a name we can use which is a little clearer?

adapters/adapters-commons/src/test/java/uk/gov/justice/raml/common/mapper/ActionMappingTest.java
```diff
@@ -73,6 +73,38 @@ public void shouldCreateMappingsCollections() throws Exception {
                 allOf(hasProperty("requestType", equalTo("application/vnd.bbbbb+json")), hasProperty("name", equalTo("actionB")))));
     }
 
+    @Test
+    public void shouldCreateMappingIfPrefixedByOtherText() throws Exception {
+
+        List<ActionMapping> mappings = listOf("test" + mappingDescriptionWith(
```
matt-rich: Rather than "test" could we use something more descriptive like "Description of action" / Action A

PR #96, opened 2016-06-15
refactor JMS sender API to facilitate JMS client generation

jaceko:
separate pull request of the refactored code to avoid big bang merge.

### DIFF
--- a/core/src/main/java/uk/gov/justice/services/core/sender/SenderProducer.java
+++ b/core/src/main/java/uk/gov/justice/services/core/sender/SenderProducer.java
@@ -8,6 +8,7 @@ import uk.gov.justice.services.core.annotation.ServiceComponent;
 import uk.gov.justice.services.core.annotation.exception.MissingAnnotationException;
 import uk.gov.justice.services.core.dispatcher.DispatcherProducer;
 import uk.gov.justice.services.core.jms.JmsSenderFactory;
+import uk.gov.justice.services.core.jms.JmsSenderWrapper;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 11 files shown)


core/src/main/java/uk/gov/justice/services/core/sender/SenderProducer.java
```diff
@@ -23,19 +24,23 @@
...
+    private ComponentDestination componentDestination;
+    private DispatcherProducer dispatcherProducer;
 
-    @Inject
-    DispatcherProducer dispatcherProducer;
+    private final Map<Component, Sender> senderMap = new ConcurrentHashMap<>();
 
-    private Map<Component, Sender> senderMap;
+    //for CDI proxying only
+    protected SenderProducer() {
```
purple52: I think we agreed a while ago to use package scoped injected fields, not constructor injection, as the lesser of all the evils to avoid adding code specifically for CDI. I think we should stick to that convention - it seems common in the JEE world.

core/src/main/java/uk/gov/justice/services/core/sender/SenderProducer.java
```diff
@@ -23,19 +24,23 @@
...
+    }
 
-    public SenderProducer() {
-        senderMap = new ConcurrentHashMap<>();
+
+    @Inject
+    public SenderProducer(final JmsSenderFactory jmsSenderFactory, final ComponentDestination componentDestination, final DispatcherProducer dispatcherProducer) {
```
purple52: See above

messaging/messaging-jms/src/main/java/uk/gov/justice/services/messaging/jms/JmsEnvelopeSender.java
```diff
@@ -13,34 +16,67 @@
...
+
+    /**
+     * Sends envelope to the destination via JMS.
+     *
+     * @param envelope    envelope to be sent.
+     * @param destinationName JNDI name of the JMS destination.
+     */
+    public void send(final JsonEnvelope envelope, final String destinationName) {
+        try {
+            Destination destination = (Destination) namingContext.lookup(destinationName);
```
purple52: final

core/src/main/java/uk/gov/justice/services/core/sender/SenderProducer.java
jaceko: If we go that way then our beans are not POJOs and can't be used outside of the container, e.g. in tests.

I've done some further investigation on bean scopes and constructors and there doesn't seem to be a good solution:
-@ApplicationScoped beans are indeed effectively singletons but require a default constructor
-by default (without any annotations) beans have a so-called dependent pseudo-scope. 
They then don't need default constructors but are not singletons. Each calling bean holds a copy of the dependent bean. Unfortunately this is also true for producers. So SenderProducer bean gets created each time a sender is produced.
-There is also a @Singleton pseudo scope which in theory does not create proxies but still for some reason requires default constructors.

PR #100, opened 2016-06-23
Dev/access control setup

allanmckenzie:


core/src/main/java/uk/gov/justice/services/core/accesscontrol/AccessControlFailureMessageGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public String errorMessageFrom(
+                    JsonEnvelope jsonEnvelope,
+                    AccessControlViolation accessControlViolation) {
+
+        final Metadata metadata = jsonEnvelope.metadata();
+        final String jsonEnvelopeName = metadata.name();
+        final String userId = metadata.userId().orElse("no user id supplied");
+
+        return format("Access Control failed for user id '%s', json envelope '%s'. Reason: %s",
```
purple52: For trace logging, we have some common components for generating JSON from the metadata of an envelope. We should use that rather than specifically pulling out the user id and name.

core/src/main/java/uk/gov/justice/services/core/accesscontrol/AccessControlService.java
```diff
@@ -0,0 +1,36 @@
...
+    public static final String ACCESS_CONTROL_DISABLED_PROPERTY =
+                    "uk.gov.justice.services.core.accesscontrol.disabled";
+
+    private final PolicyEvaluator policyEvaluator;
+
+    public AccessControlService() {
+        this(new AllowAllPolicyEvaluator());
+    }
+
+    public AccessControlService(PolicyEvaluator policyEvaluator) {
```
purple52: We should use our CDI convention here - the PolicyEvaluator should be package scoped and injected with @Inject, and no constructor. We will create a PolicyEvaluatorProducer to inject it, I think.

core/src/main/java/uk/gov/justice/services/core/dispatcher/Dispatcher.java
```diff
@@ -64,6 +78,11 @@ void register(final Object handler) {
         handlerRegistry.register(handler);
     }
 
+    private JsonEnvelope doDispatch(final JsonEnvelope envelope, final boolean isSynchronous) {
+        checkAccessControl(envelope);
+        return (JsonEnvelope) getMethod(envelope, isSynchronous).execute(envelope);
```
purple52: I think we should have the "if" statement here, so we're not using the exception to control flow.

PR #102, opened 2016-06-27
Refactor DispatcherProducer into component parts

mapingo:


core/src/main/java/uk/gov/justice/services/core/dispatcher/Dispatcher.java
```diff
@@ -38,7 +34,7 @@
      *
      * @param envelope the envelope to dispatch to a handler
      */
-    void asynchronousDispatch(final JsonEnvelope envelope) {
+    public void asynchronousDispatch(final JsonEnvelope envelope) {
```
purple52: If this becomes public, then so should synchronousDispatch.

core/src/main/java/uk/gov/justice/services/core/dispatcher/RequesterProducer.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.justice.services.core.dispatcher;
+
+import uk.gov.justice.services.core.annotation.ServiceComponent;
+
+import javax.enterprise.inject.Produces;
+import javax.enterprise.inject.spi.InjectionPoint;
+import javax.inject.Inject;
+
+public class RequesterProducer {
```
purple52: Needs to be @ApplicationScoped

core/src/main/java/uk/gov/justice/services/core/dispatcher/Dispatcher.java
allanmckenzie: good point

mapingo: synchronousDispatch is now public, this can be reverted when SenderProducer is removed/refactored.

PR #106, opened 2016-06-30
Add enevlope builder

jaceko:


messaging/messaging-core/src/main/java/uk/gov/justice/services/messaging/DefaultJsonEnvelope.java
```diff
@@ -54,5 +58,50 @@ public JsonString payloadAsJsonString() {
...
+    }
+
+    public static class Builder {
+        private JsonObjectMetadata.Builder metadata;
+        private final JsonObjectBuilderWrapper payload = new JsonObjectBuilderWrapper();
+
+        private Builder() {
+        }
+
+        public Builder with(final JsonObjectMetadata.Builder metadata) {
```
purple52: Do we want to call this withMetadata or withMetadataOf?

jaceko: typically you would be saying be saying
with(metadataOf(...) ... ) using the static metadataOf method  from the JsonObjectMetadata class. You get more fluent API that way.

purple52: Ah, makes sense!

PR #107, opened 2016-07-01
Add EVENT_API component

skiddykong:


core/src/main/java/uk/gov/justice/services/core/annotation/Component.java
```diff
@@ -26,6 +26,7 @@
     COMMAND_HANDLER("command", "handler", Queue.class),
     EVENT_LISTENER("event", "listener", Topic.class),
     EVENT_PROCESSOR("event", "processor", Topic.class),
+    EVENT_API("event", "api", Queue.class),
```
matt-rich: I wasn't part of this discussion, but how are we sending things out of the API layer? We don't have an standard event Handler/Controller control flow to send a JMS message to a single Queue... Is it possible that BOTH the Event listener and processor tiers would need to receive a message created by the event-api? - in which case would this be a topic? - Happy for you guys to discuss and then just fill me in, on slack or on Monday :)

purple52: I think it should be null - in this implementation the enum defines what it's listening to, and the event api will not be listening to JMS

PR #109, opened 2016-07-05
Add custom component annotation

jaceko:


core/src/main/java/uk/gov/justice/services/core/annotation/FrameworkComponent.java
```diff
@@ -0,0 +1,21 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import javax.enterprise.context.ApplicationScoped;
+
+/**
+ * Identifies framework components. <p> Usage: @ServiceComponent("COMPONENT_NAME")
```
matt-rich: Usage here in description is incorrect.

jaceko: fixed

PR #117, opened 2016-07-08
 Allow event-api component base URI in RAML rest generator

skiddykong:


generators/rest-adapter-generator/src/test/java/uk/gov/justice/services/adapters/rest/generator/GeneratorsTest.java
```diff
@@ -4,6 +4,7 @@
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
+import static uk.gov.justice.services.core.annotation.Component.*;
```
purple52: Remove wildcard import

PR #128, opened 2016-07-25
Add query handler testing utils and apply to example service

mapingo:


jaceko:
Good stuff, would be good to have that for other types of handlers.
Also we assume here that query handler always calls the requester, which is not always the case.

PR #130, opened 2016-07-28
Relax messaging resource uri validation

jaceko:


generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
```diff
@@ -210,25 +209,17 @@ private AnnotationSpec generateActivationConfigPropertyAnnotation(final String n
      * @return camel case class name
      */
     private String classNameOf(final MessagingResourceUri resourceUri, final Component component) {
```
mapingo: component parameter can be removed as it is unused

jaceko: fixed

PR #137, opened 2016-08-08
add convenience classes to framework

allanmckenzie:


test-utils/src/main/java/uk/gov/justice/services/test/utils/messaging/MessageTestClient.java
```diff
@@ -0,0 +1,82 @@
...
+    public void close() {
+        close(messageConsumer);
+        close(session);
+
+        session = null;
+        topic = null;
+        messageConsumer = null;
+    }
+
+    private void close(AutoCloseable closeable) {
```
mapingo: closeable could be final

PR #142, opened 2016-08-11
Add dispatcher interceptor implementation

mapingo:


core/src/main/java/uk/gov/justice/services/core/annotation/FrameworkInterceptor.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Identifies interceptor classes, which are used to intercept incoming JsonEnvelopes before the
+ * dispatcher, or out going messages before they are returned to the adapter.  The class must
+ * implement {@link Interceptor}
+ */
+@Retention(RUNTIME)
+@Target(TYPE)
+@ApplicationScoped
+public @interface FrameworkInterceptor {
```
matt-rich: If all classes that have this annotation also have to implement the Interceptor interface then isn't there an easier way of just getting all beans that implement that and using that rather than the annotation injection?

You should be able to intercept the after deployment validation event and find all beans of Type Interceptor ?

core/src/main/java/uk/gov/justice/services/core/extension/AnnotationScanner.java
```diff
@@ -41,15 +43,21 @@
 
     @SuppressWarnings("unused")
     void afterDeploymentValidation(@Observes final AfterDeploymentValidation event, final BeanManager beanManager) {
-        beanManager.getBeans(Object.class, annotationLiteral()).stream()
+        final Set<Bean<?>> beans = beanManager.getBeans(Object.class, annotationLiteral());
```
matt-rich: Rather than Object.class can you just get those for your Interceptor.class interface? - shortcutting the process, or maybe something similar to this, see my comment on the new annotation.

Also I think it was suggested that perhaps if we were to start adding even more annotations we should refactor this code to make it cleaner rather than running through the streams multiple times.

core/src/main/java/uk/gov/justice/services/core/extension/FrameworkInterceptorFoundEvent.java
```diff
@@ -0,0 +1,22 @@
...
+import uk.gov.justice.services.core.annotation.FrameworkInterceptor;
+
+import javax.enterprise.inject.spi.Bean;
+
+/**
+ * Event representing the occurrence of a class with an {@link FrameworkInterceptor} annotation
+ * having been identified by the framework.
+ */
+public class FrameworkInterceptorFoundEvent {
+
```
matt-rich: See earlier comments, this may not be required.

core/src/main/java/uk/gov/justice/services/core/interceptor/AsynchronousDispatchInterceptor.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.justice.services.core.interceptor;
+
+import uk.gov.justice.services.messaging.JsonEnvelope;
+
+import java.util.function.Consumer;
+
+/**
+ * Interceptor dispatch target for asynchronous calls to dispatch.
+ */
+public class AsynchronousDispatchInterceptor implements Interceptor {
```
matt-rich: Should this have an annotation ?

core/src/main/java/uk/gov/justice/services/core/interceptor/AsynchronousDispatchInterceptor.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Calls the dispatcher with the input JsonEnvelope from the interceptor context.
+     *
+     * @param interceptorContext the context that contains the input JsonEnvelope
+     * @param interceptorChain   the interceptor chain, not used as this is the target of the chain
+     * @return the interceptor context passed to this method as there is no output from the
+     * dispatcher
+     */
+    @Override
+    public InterceptorContext process(final InterceptorContext interceptorContext, final InterceptorChain interceptorChain) {
```
matt-rich: Having a not used parameter for a method implemented through an 'interface' contract doesn't have a good smell about it...

core/src/main/java/uk/gov/justice/services/core/interceptor/ChainProcess.java
```diff
@@ -0,0 +1,13 @@
...
+import uk.gov.justice.services.messaging.JsonEnvelope;
+
+/**
+ * Functional interface for the chain process this represents both asynchronous and synchronous
+ * function calls.
+ */
+@FunctionalInterface
+public interface ChainProcess {
+
+    JsonEnvelope process(final JsonEnvelope envelope);
```
matt-rich: Are we processing a JsonEnvelope or something wrapping it?

core/src/main/java/uk/gov/justice/services/core/extension/BeanInstantiater.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.services.core.extension;
+
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.inject.Inject;
+
+public class BeanInstantiater {
```
matt-rich: May want a class comment here.

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorChainProcessorProducer.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Produces the correct implementation of an synchronous chain process depending on the {@link
+     * Adapter} annotation at the injection point.
+     *
+     * @param injectionPoint class where the {@link InterceptorChainProcessor} is being injected
+     * @return the chain process
+     */
+    @Produces
+    public InterceptorChainProcessor produceProcessor(final InjectionPoint injectionPoint) {
+        //TODO: As part of integration into the framework the dispatcher needs to be simplified so access is through a single method
```
matt-rich: TODO comment, is this part of the later changes?

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorContext.java
```diff
@@ -0,0 +1,100 @@
...
+    private final ContextPayload output;
+    private final InjectionPoint injectionPoint;
+
+    private InterceptorContext(final ContextPayload input, final ContextPayload output, final InjectionPoint injectionPoint) {
+        this.input = input;
+        this.output = output;
+        this.injectionPoint = injectionPoint;
+    }
+
+    private InterceptorContext(final JsonEnvelope input, final InjectionPoint injectionPoint) {
```
matt-rich: I know this is a private method, but do you think it is worth documenting that without output specified it uses the input as an output? It is kind of a fundamental part?

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorObserver.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.enterprise.inject.spi.AfterDeploymentValidation;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.Extension;
+
+/**
+ * Observes for {@link AfterDeploymentValidation} and adds all {@link Interceptor} implementations
+ * to the {@link InterceptorCache}
+ */
+public class InterceptorObserver implements Extension {
```
matt-rich: If we are having a specific observer for Interceptor.class would it be worth putting this behind a pattern that all of our observers use? - just a consideration if this is preferred standard way to do this.

Even potentially templating the class to be used based on the implementation?

core/src/test/java/uk/gov/justice/services/core/interceptor/InterceptorContextTest.java
```diff
@@ -0,0 +1,97 @@
...
+import javax.enterprise.inject.spi.InjectionPoint;
+
+import org.junit.Test;
+
+public class InterceptorContextTest {
+
+    @Test
+    public void shouldBuildInterceptorContextWithInputAndInjectionPoint() throws Exception {
+
+        final JsonEnvelope input = mock(JsonEnvelope.class);
```
matt-rich: could we make some of these field level mocks to reduce duplication?

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorObserver.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private final List<Bean<?>> interceptorBeans = new ArrayList<>();
+
+    @SuppressWarnings({"unchecked", "unused"})
+    void afterDeploymentValidation(@Observes final AfterDeploymentValidation event, final BeanManager beanManager) {
```
matt-rich: Could be nice to use a peek here to add each bean to a log entry as well so we know what Interceptors are being logged.

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorChainProcessorProducer.java
mapingo: Yes, this is completed when the integration changes are made.

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorContext.java
mapingo: I have added some more java docs

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorObserver.java
mapingo: If you are happy with the current implementation, can we merge for now and discuss the best practice for this, next week

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorObserver.java
mapingo: done

core/src/main/java/uk/gov/justice/services/core/interceptor/InterceptorObserver.java
matt-rich: Sure

PR #164, opened 2016-08-26
Fix remote Client unable to post to rest end point

mapingo:


generators/rest-client-generator/src/test/java/uk/gov/justice/services/clients/rest/generator/RestClientGenerator_CodeStructureTest.java
```diff
@@ -246,4 +239,8 @@ private void assertRestClientHelperField(Field field) {
         assertThat(field.getAnnotation(Inject.class), not(nullValue()));
     }
 
+    private void assertEnveloperField(Field field) {
```
matt-rich: final

PR #177, opened 2016-09-07
comment the toDebugStringPrettyPrint() method on JsonEnvelope in the …

allanmckenzie:
…hope that people should use it correctly


messaging/messaging-core/src/main/java/uk/gov/justice/services/messaging/JsonEnvelope.java
```diff
@@ -21,6 +21,27 @@
...
+     * N.B. This should NEVER be used in production code as the json
+     * may contain sensitive data which can be harmful should it appear
+     * in logs.
+     *
+     * If you want to log this envelope for tracing purposes
+     * then use the toString() method, as this will give the
+     * objects metadata without exposing anything sensistive
+     *
+     * @return this envelope as it's json string, pretty printed.
+     **/
```
matt-rich: Mark as deprecated?

PR #184, opened 2016-09-12
Implement sender.sendAsAdmin()

jaceko:


generators/messaging-client-generator/pom.xml
```diff
@@ -45,6 +45,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>uk.gov.justice.services</groupId>
+            <artifactId>test-utils-common</artifactId>
+            <version>${project.version}</version>
```
purple52: SHould this be test scope?

jaceko: that's actually not needed, removing it...

generators/messaging-client-generator/src/test/java/uk/gov/justice/test/util/RecordingJmsEnvelopeSender.java
```diff
@@ -25,6 +25,7 @@ public void init() {
 
     @Override
     public void send(final JsonEnvelope envelope, final Destination destination) {
+        System.out.println("AAAAAAAAAAAAAAAAA");
```
allanmckenzie: Is this intentional?

PR #189, opened 2016-09-16
add a varargs method to the database cleaner

allanmckenzie:


test-utils/test-utils-persistence/src/main/java/uk/gov/justice/services/test/utils/persistence/DatabaseCleaner.java
```diff
@@ -46,6 +50,16 @@ public void cleanEventLogTable(final String contextName) {
         }
     }
 
+    public void cleanViewStoreTables(final String contextName, final String tableName, final String... tableNames) {
```
matt-rich: Is there any reason why we still need the 2nd 'tableName' as a second argument?

It isn't very clear why it can't be rolled into the vararg

allanmckenzie: So that there must be at least one table name

matt-rich: ok, should the var name be additionaTableNames or similar? or with a comment?

allanmckenzie: done :-)

PR #190, opened 2016-09-19
Add strategy for buffer locking

jaceko:


event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/service/BufferInitialisationStrategyProducer.java
```diff
@@ -0,0 +1,29 @@
...
+    @Inject
+    @Value(key = "postgres.db", defaultValue = "true")
+    String usePostgresStrategy;
+
+    @Inject
+    StreamStatusJdbcRepository streamStatusRepository;
+
+    @Produces
+    public BufferInitialisationStrategy bufferInitialisationStrategy() {
+        return TRUE.equals(usePostgresStrategy)
```
matt-rich: Can you do Boolean.valueOf() for if it is "true" ?

event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/service/BufferInitialisationStrategyProducer.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.enterprise.inject.Produces;
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class BufferInitialisationStrategyProducer {
+
+    private static final String TRUE = "true";
+    @Inject
+    @Value(key = "postgres.db", defaultValue = "true")
+    String usePostgresStrategy;
```
purple52: I suggest we set the property to be a fully qualified class name. Then any strategy can be added via the classpath. If no value is provided, it can default to the ANSI strategy. Also suggest giving the property name more of a namespace, so it is clear where the property is used.

jaceko: OK but do we really want to default to the ANSI strategy? 
That way we'd have to make changes in the vagrant box (for each context). I'd say let's make the Postgres strategy default then no changes are required

purple52: Yes, the Wildfly config would need to change - I think selecting a non-standard strategy should be an explicit thing. This would need to be communicated to devops.

jaceko: Implemented the producer picking up fully qualified strategy name.

event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/service/BufferInitialisationStrategyProducer.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.slf4j.Logger;
+
+@ApplicationScoped
+public class BufferInitialisationStrategyProducer {
+
+    private static final String INSTANTIATION_ERROR_MSG = "Could not instantiate buffer initialisation strategy.";
+
+    @Inject
+    @GlobalValue(key = "event.buffer.init.strategy", defaultValue = "uk.gov.justice.services.event.buffer.core.service.PostgreSQLBasedBufferInitialisationStrategy")
```
matt-rich: I really don't like this, but this is what the chosen strategy was?

jaceko: it's implemented as per Davids and Andrews comments.
the property was to be a fully qualified class name (David) and the default strategy was to be the PostgreSQL one (Andrew)

event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/service/BufferInitialisationStrategyProducer.java
```diff
@@ -0,0 +1,42 @@
...
+    Logger logger;
+
+    @Inject
+    StreamStatusJdbcRepository streamStatusRepository;
+
+    @Produces
+    public BufferInitialisationStrategy bufferInitialisationStrategy() {
+        logger.info("Instantiating {}", strategyClass);
+        try {
+            Class<?> clazz = Class.forName(strategyClass);
```
matt-rich: Should we check the injected class is an instance of BufferInitialisationStrategy ?

jaceko: That way or another, we should throw classCastException, so no reason to add extra ifs, IMHO

event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/service/BufferInitialisationStrategyProducer.java
matt-rich: Yea that is what I thought, if we refactor the code we won't realise the package is wrong... but fair enough.

event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/service/BufferInitialisationStrategyProducer.java
matt-rich: ok, just wondered if we would wrap it in a nice exception as opposed to just a generic class cast exception.. but yea fine. Will merge.

PR #193, opened 2016-09-20
Improve database cleaner

allanmckenzie:


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/http/PollingRestClient.java
```diff
@@ -0,0 +1,122 @@
...
+    private final Sleeper sleeper;
+
+    /**
+     * Constructs a fully functioning PollingRestClient
+     */
+    public PollingRestClient() {
+        this(new ValidatingRestClient(), new Sleeper());
+    }
+
+    public PollingRestClient(final ValidatingRestClient validatingRestClient, Sleeper sleeper) {
```
mapingo: Missing final for Sleeper

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/http/HttpResponsePoller.java
```diff
@@ -20,6 +20,10 @@
 import javax.json.JsonReader;
 import javax.ws.rs.core.Response;
 
+/**
+ *
+ */
+@Deprecated
```
mapingo: Do we want to deprecate at the moment?  If we do then this needs a comment of where to go for the new version.

PR #194, opened 2016-09-20
Extend DefaultJsonEnvelope to accept JsonObject as part of the payload

deepakhalale:


deepakhalale:
The above changes are only intended to be used in test cases. In the notification context that I am working on, it's required to create a JsonEnvelope with JsonObject in the payload, for which I am using the Builder from DefaultJsonEnvelope.

deepakhalale:
Have added a test case as per the review comments from Shaun.

deepakhalale:
After discussing with @matt-rich it turns out that I don't need to update DefaultJsonEnvelope for what I am trying to test in notification context. Also updating production code which is used for only testing purpose doesn't seem right. Hence closing the PR.

PR #200, opened 2016-09-28
Sibyg

sibyg:
Review comments incorporated


common/src/main/java/uk/gov/justice/services/common/converter/LocalDateTimes.java
```diff
@@ -0,0 +1,39 @@
+package uk.gov.justice.services.common.converter;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Utility functions for converting to and from {@link LocalDate} objects
+ * and ISO date format String such as '2011-12-03'
+ */
+public final class LocalDateTimes {
```
purple52: This should be called LocalDates - it doesn't deal with LocalDateTime.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/IntegerGenerator.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.services.test.utils.core.random;
+
+class IntegerGenerator extends Generator<Integer> {
```
purple52: Shouldn't this be public?

PR #203, opened 2016-09-29
TypeCheck utility added

sibyg:


sibyg:
forked from an old branch

PR #204, opened 2016-09-29
Add Utility class TypeCheck for type-based testing

sibyg:
It supports n times condition checking for random values of a type T.

### DIFF
--- a/test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/helper/TypeCheck.java
+++ b/test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/helper/TypeCheck.java
@@ -1,0 +1,70 @@
+package uk.gov.justice.services.test.utils.core.helper;
+
+import uk.gov.justice.services.test.utils.core.random.Generator;
+
+import java.util.function.Function;
+
+import static java.util.Objects.requireNonNull;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+/**
+ * A Utility class for type-based testing, where a type is a high-level specification of behavior that should hold for a range of data point
+ *
+ * @param <T> Type of the Property and must have {@link  Generator} implementation for type T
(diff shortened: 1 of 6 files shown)


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/helper/TypeCheck.java
```diff
@@ -0,0 +1,56 @@
...
+    private int numberOfTimes;
+
+    private TypeCheck(final Generator<T> generator, final Function<T, Boolean> condition) {
+        requireNonNull(generator, "Generator cannot be null");
+        requireNonNull(condition, "Condition cannot be null");
+        this.generator = generator;
+        this.condition = condition;
+    }
+
+    public static <T> TypeCheck<T> aTypeCheck(final Generator<T> generator, final Function<T, Boolean> condition) {
```
purple52: Our convention would be to call this "typeCheck" - we don't put "a" or "an" at the start of factory methods.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/helper/TypeCheck.java
```diff
@@ -0,0 +1,56 @@
...
+        return new TypeCheck<T>(generator, condition);
+    }
+
+    public TypeCheck<T> numberOfTimes(final int numberOfTimes) {
+        this.numberOfTimes = numberOfTimes;
+        return this;
+    }
+
+    public TypeCheck<T> preCondition(final Function<T, Boolean> preCondition) {
```
purple52: withPrecondition - precondition is one word, and "with" at the start indicates this is a builder method

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/helper/TypeCheck.java
```diff
@@ -0,0 +1,56 @@
...
+        this.preCondition = preCondition;
+        return this;
+    }
+
+    public void verify() {
```
purple52: You could follow the Mockito model for this and pass a number of times to the verify method, given that the number of times is a required field. eg, :
verify() (would run once)
verify(times(20)) (would verify 20 times)

PR #212, opened 2016-10-04
Add JsonEnvelope Hamcrest matchers

mapingo:


test-utils/test-utils-core/pom.xml
```diff
@@ -60,6 +60,15 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.jayway.jsonpath</groupId>
+            <artifactId>json-path-assert</artifactId>
```
allanmckenzie: test scope?

PR #213, opened 2016-10-05
fix intergration test port 8080 problem

allanmckenzie:


generators/rest-client-core/src/main/java/uk/gov/justice/services/clients/core/RestClientProcessor.java
```diff
@@ -193,13 +193,16 @@ private WebTarget createWebTarget(final EndpointDefinition definition, final Jso
     }
 
     private String createBaseUri(final EndpointDefinition definition) {
-        // TODO: Port and same service check and temporary solutions
-        return isEmpty(port) || isSameService(definition) ? definition.getBaseUri() : definition.getBaseUri().replace(":8080", ":" + port);
```
matt-rich: So we will never use the port in the base URI now?

PR #218, opened 2016-10-06
Add Artemis dependency and share subscriptions

gogit:
Add Artemis dependency and share subscriptions config property.

### DIFF
--- a/generators/messaging-adapter-generator/pom.xml
+++ b/generators/messaging-adapter-generator/pom.xml
@@ -3,12 +3,11 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-        <artifactId>generators</artifactId>
         <groupId>uk.gov.justice.services</groupId>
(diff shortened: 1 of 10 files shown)


generators/messaging-adapter-generator/pom.xml
```diff
@@ -107,6 +125,34 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-ra</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>uk.gov.justice.artemis</groupId>
+            <artifactId>embedded-artemis</artifactId>
+            <version>1.0.0</version>
```
purple52: Can we pull this version number out into a property, and a dependency management definition at the top level of the project?

generators/messaging-adapter-generator/src/main/resources/org/apache/openejb/service-jar.xml
```diff
@@ -0,0 +1,846 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gogit: Moved the files to test resources folder.

PR #219, opened 2016-10-07
Fix event buffer connection held open after exception issue

mapingo:


components/event/event-listener/src/main/java/uk/gov/justice/services/event/buffer/EventBufferInterceptor.java
```diff
@@ -37,8 +39,11 @@ public InterceptorContext process(final InterceptorContext interceptorContext, f
     }
 
     private Stream<InterceptorContext> streamFromEventBufferFor(final InterceptorContext interceptorContext) {
-        return eventBufferService.currentOrderedEventsWith(interceptorContext.inputEnvelope())
-                .map(jsonEnvelope -> copyWithInput(interceptorContext, jsonEnvelope));
+        final Stream<JsonEnvelope> envelopeStream = eventBufferService.currentOrderedEventsWith(interceptorContext.inputEnvelope());
+        final List<JsonEnvelope> envelopes = envelopeStream.collect(Collectors.toList());
```
matt-rich: static import?

components/event/event-listener/src/main/java/uk/gov/justice/services/event/buffer/EventBufferInterceptor.java
```diff
@@ -37,8 +39,11 @@ public InterceptorContext process(final InterceptorContext interceptorContext, f
...
+        envelopeStream.close();
```
matt-rich: if an exception occurs is this still skipped?

components/event/event-listener/src/main/java/uk/gov/justice/services/event/buffer/EventBufferInterceptor.java
```diff
@@ -37,8 +39,11 @@ public InterceptorContext process(final InterceptorContext interceptorContext, f
     }
 
     private Stream<InterceptorContext> streamFromEventBufferFor(final InterceptorContext interceptorContext) {
```
matt-rich: Add changeling entry.

components/event/event-listener/src/main/java/uk/gov/justice/services/event/buffer/EventBufferInterceptor.java
mapingo: This code is run prior to any possible exception being thrown in the Handles methods...so connections will always be closed before continuing with the interceptor chain

matt-rich: ok, changes look good then.

PR #223, opened 2016-10-10
Extend DefaultJsonEnvelope.Builder and update random generators with EnumPicker

deepakhalale:


jaceko:
This is an old builder, we should delete it and use DefaultJsonEnvelope.Builder

PR #225, opened 2016-10-10
Deprecate JsonEnvelopeBuilder

purple52:
Use DefaultEnvelopeBuilder.envelope() instead


deepakhalale:
#223 has changes to Extend DefaultJsonEnvelope.Builder to accept JsonObjectBuilder as part of the payload which hasn't been replicated in this PR. Hence keeping in intact in the PR.

PR #232, opened 2016-10-14
Improve test utils for passthrough handler methods.

akomisarek:
* Add method name in front of sender invocation.
* Add method name in assertion message if the requester response is wrong
* Add method name in front of requester invocation.


akomisarek:
Commit message amended.

PR #235, opened 2016-10-19
Filter out invalid character for class name  when generating class

y725:


generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/helper/MessagingResourceUri.java
```diff
@@ -26,7 +26,16 @@ public String destinationName() {
      * @return capitalised uri string without delimiters
      */
     public String toCapitalisedString() {
```
purple52: This method probably needs to be renamed if its only being used to generate a class name. Maybe split out the logic that generates a name from the logic that ensures its a valid class name?

generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/helper/MessagingResourceUri.java
```diff
@@ -26,7 +26,16 @@ public String destinationName() {
...
-        return capitalize(stripStart(uriString, "/"), '.').replace(".", "");
+        //Remove any character that is invalid as starting character for a class name
+        String uriStringStriped = uriString.replaceAll("^[^A-Za-z_$]+", "");
+
+        final StringBuilder nonWordCharAsString = new StringBuilder();
+        uriString.chars().filter(e -> !Character.isLetter(e)).forEach(e -> nonWordCharAsString.append((char) e));
+        char[] delimiters = nonWordCharAsString.toString().toCharArray();
+        String capitalizedWithCharacter = capitalize(uriStringStriped, delimiters);
```
purple52: This is a very round-the-houses bit of code. I suggest simplifying as follows:
cleaned = orig.replaceAll("[^A-Za-z0-9_$]", " ")  - ie, replace all other chars with a space
captialised = capitaliseFully(cleaned, {' ', '_', '$'}) - you have a fixed set of delimiters now so this is simple
result = capitalised.replaceAll(" ", "") - remove the spaces
then remove the 1st char if it is a number, possibly in a separate method
Much of this could be inline to avoid variable names for partial results that don't really mean anything

liyon: That looks much better but need also add numbers to the delimiters otherwise it will not work.
so the second line need to be:
     captialised = capitaliseFully(cleaned, " _$0123456789".toCharArray());

PR #240, opened 2016-10-25
Add proxy support to RestClient

deepakhalale:


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/rest/RestClient.java
```diff
@@ -2,18 +2,26 @@
...
 import javax.ws.rs.core.Response;
 
+import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
+
 /**
  * Helper class to send Post Commands and Get Queries.
  */
 public class RestClient {
 
+    private final ResteasyClientBuilder clientBuilder;
```
allanmckenzie: Do you need to store the client builder as a member variable? Surely as it's created using a static method it could be created anew each time it's needed

deepakhalale: Updated to create a new each time it's used.

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/rest/ResteasyClientBuilderFactoryTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+        assertThat(actualProxySettings.getHostName(), is(PROXY_HOST));
+        assertThat(actualProxySettings.getPort(), is(PROXY_PORT));
+
+        clearProxyConfiguration();
+    }
+
+    @Test
+    public void shouldNotSetProxyIfNotConfigured() throws Exception {
+        givenProxyIsNotConfigured();
```
purple52: Why are we testing that no proxy is set? We might want to ensure that before the test is run that the properties are clear, but it's entirely possible that the tests might be run on an environment where the system properties **are** set, because they are genuinely needed.

If these tests need the values to be clear, I suggest you grab the existing values before the test, clear them, then set them back at the end.

PR #245, opened 2016-10-27
add new finder method to dispatcher cache

allanmckenzie:


core/src/main/java/uk/gov/justice/services/core/dispatcher/DispatcherCache.java
```diff
@@ -49,6 +50,10 @@ public Dispatcher dispatcherFor(final ServiceComponentFoundEvent event) {
                 event.getComponentName(), event.getLocation()));
     }
 
+    public Dispatcher dispatcherFor(final Component component, final ServiceComponentLocation location) {
```
mapingo: Would be nice to have a java doc for this

PR #248, opened 2016-10-31
Refactor random email generator

gogit:
Refactor email generator


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+ * <li>Upper case and lower case Latin letters A to Z and a to z</li>
+ * <li>digits 0 to 9, provided that top-level domain names are not all-numeric</li>
+ * <li>hyphen -, provided that it is not the first or last character</li>
+ * </ol>
+ * 
+ * Sources: <br>
+ * https://tools.ietf.org/html/rfc3696 <br>
+ * https://en.wikipedia.org/wiki/Email_address#domain
+ * 
+ * @return
```
deepakhalale: @return annotation not required.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public class DomainPartGenerator {
+
+    /**
+     * The maximum length of the domain part
+     */
+    public static final int MAX_LENGTH = 63;
```
deepakhalale: can be access level not be private. If used in testing, it needs to have it's own constants as it's part of the contract what we are testing.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+    public static final char[] DOMAINPART_CHARACTERS =
+                    "-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789".toCharArray();
+
+
+
+    /**
+     * Version 2016102700, Last Updated Thu Oct 27 07:07:01 2016 UTC <br>
+     * http://data.iana.org/TLD/tlds-alpha-by-domain.txt
+     */
+    private final List<String> tlds = new ArrayList<>();
```
deepakhalale: Can the variable name be more descriptive. May be what TLD stands for?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Random used for character generation
+     */
+    private final Random random;
+
+    /**
+     * Prevent instantiation outside the package
+     * 
+     * @throws IOException
```
deepakhalale: Misleading exception is java doc as the code seems to do catch all.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    DomainPartGenerator(final Random random) {
+        try {
+            final Properties props = new Properties();
+            props.load(getClass().getClassLoader().getResourceAsStream("tld.txt"));
+            for (final Object key : props.keySet()) {
```
deepakhalale: Do we have Java 8 equivalent for the loop

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+                tlds.add(String.valueOf(key).trim());
+            }
+        } catch (Exception e) {
+            // fall back to some tlds
+            tlds.add("gov.uk");
```
deepakhalale: Can we squash this to one liner with addAll may be

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * The maximum length of the domain part
+     */
+    public static final int MAX_LENGTH = 63;
+
+    /**
+     * List of valid characters in the domain part of the email <br>
+     * LDH rule ( letters digits and the hyphen )
+     */
+    public static final char[] DOMAINPART_CHARACTERS =
```
gogit: Part of the contract exposed by the Domain Part , I'll add an additional test.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
gogit: fairly common terminology, same as URL ?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
gogit: changing

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
gogit: changing

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Prevent instantiation outside the package
+     * 
+     * @throws IOException
+     */
+    DomainPartGenerator(final Random random) {
+        try {
+            final Properties props = new Properties();
+            props.load(getClass().getClassLoader().getResourceAsStream("tld.txt"));
```
gogit: we probably should not be using external libraries if we can do the same thing via jdk libraries ?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
gogit: would make it harder to read ? if the number of entries increase.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
deepakhalale: OK a test case checking for length would do as well. I idea is if someone changes the length to say 64 in the code, the test cases would still work as they share the same variable!

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
deepakhalale: This is to be consistent with rest of the implementation in the framework and I don't see any harm in using the libraries if we have already included them in the project.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
deepakhalale: It could look like e.g.
tlds.addAll(newArrayList("gov.uk", "co.uk", "org", "something", "something-else", "uk"));
which is concise and can handle N-number of elements.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
gogit: consistency is not an argument  for using external libraries :), we  can always refactor and reduce the number of external libraries

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
gogit: changing

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
gogit: creating objects when not needed is not good practice

deepakhalale: well I disagree in this case. But it's upto you.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+                generated = String.join("", generated, "(comment)");
+            }
+        }
+        generated = String.join(".", generated, tlds.get(chooseRandomPosition(tlds.size())));
+
+        return generated;
+    }
+
+
+    public boolean checkPart(final String textToCheck) {
```
deepakhalale: Is checkPart used outside the class as it seems to have public visibility?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
allanmckenzie: I agree with Deepak - name needs to be more descriptive. Unless you can be sure that all developers both now and in the future will understand it, don't use acronyms

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtil.java
```diff
@@ -0,0 +1,103 @@
...
+     * Check that the text contains only the valid characters mentioned
+     * 
+     * @param text to check
+     * @param valid characters
+     * @return flag indicating that the check passed
+     */
+    public static boolean checkValidityOfText(final String text, final char[] valid) {
+        final char[] local = text.toCharArray();
+        boolean flag = false;
+        for (char c : local) {
```
allanmckenzie: Not sure about the logic here. What happens when the first character is invalid but all subsequent characters are valid. Wouldn't it still return true?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtil.java
```diff
@@ -0,0 +1,103 @@
...
+     * Choose from a given length <br>
+     * If the length is n, returns approximately 1 in n as positive <br>
+     * If the length is 6, returns approximately 1 in 6 as positive <br>
+     * If the length is 2, returns approximately 1 in 2 as positive
+     * 
+     * @param length
+     * @return flag
+     */
+    public static boolean chooseRandom(final int length) {
+        return (new Random().nextInt(length) == 0) ? true : false;
```
allanmckenzie: This seems to be a bit of an unusual way of doing this. Wouldn't 

random.nextInt() % length 

be better?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtil.java
```diff
@@ -0,0 +1,103 @@
...
+     * 
+     * Choose from a given length <br>
+     * If the length is n, returns approximately 1 in n as positive <br>
+     * If the length is 6, returns approximately 1 in 6 as positive <br>
+     * If the length is 2, returns approximately 1 in 2 as positive
+     * 
+     * @param length
+     * @return flag
+     */
+    public static boolean chooseRandom(final int length) {
```
allanmckenzie: Is this a good name? Maybe something like onceOutOf(int denominator). 

Which would read 

boolean isTrue = onceOutOf(6)

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+     * Generate domain part
+     * 
+     * @return domain part
+     */
+    private String generateDomainPart() {
+        String generated = generate(random, DOMAINPART_CHARACTERS, 1, MAX_LENGTH);
+        if (!checkPart(generated)) {
+            return generateDomainPart();
+        }
+        // toss 100 sided thing
```
matt-rich: are these comments needed?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+        if (chooseRandom(100)) {
+            // toss coin
+            if (chooseRandom(2)) {
+                generated = String.join("", "(comment)", generated);
```
matt-rich: Is using StringJoiner (merge method maybe) cleaner here?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/DomainPartGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+
+
+    /**
+     * Check that the text is valid
+     * 
+     * @param textToCheck
+     * @return boolean value denoting a valid or invalid domain part
+     */
+    public boolean checkPart(final String textToCheck) {
+        return !textToCheck.isEmpty() && !textToCheck.startsWith("-") && !textToCheck.endsWith("-")
```
matt-rich: can we replace this with a regex pattern matcher.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/LocalPartGenerator.java
```diff
@@ -0,0 +1,106 @@
...
+        return generateLocalPart();
+    }
+
+    /**
+     * Check the local part of the email
+     * 
+     * @param textToCheck
+     * @return flag indicating check passed
+     */
+    @VisibleForTesting
```
allanmckenzie: Maybe extract to it's own class for testing rather than making it visible. Seems a shame to change your production code to make testing easier

gogit: well, conflicting opinions on what needs to be done :)

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/DomainPartGeneratorTest.java
```diff
@@ -0,0 +1,78 @@
...
+        DomainPartGenerator dpg = new DomainPartGenerator(new Random());
+        for (int i = 0; i < 10000; i++) {
+            String domainPart = dpg.next();
+            assertTrue(checkValidityOfText(domainPart, valid));
+        }
+    }
+
+    @Test
+    public void shouldFlagDomainPartStartingOrEndingWithHyphenAsInvalid() {
+        DomainPartGenerator dpg = new DomainPartGenerator(new Random());
```
allanmckenzie: final

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/EmailAddressGeneratorTest.java
```diff
@@ -0,0 +1,27 @@
...
+public class EmailAddressGeneratorTest {
+
+    @Test
+    public void shouldCheckConstructorIsNotAccessible()
+                    throws NoSuchMethodException, SecurityException {
+        Constructor<EmailAddressGenerator> c = EmailAddressGenerator.class.getDeclaredConstructor();
+        assertFalse(c.isAccessible());
+    }
+
+    @Test
```
allanmckenzie: Not really a test. Better to use a regular expression and check each is unique

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/LocalPartGenerator.java
purple52: I'm with @allanmckenzie - I don't like private methods being exposed using for testing.

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtilTest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Test
+    public void shouldTestThatStringIsComprisedOfValidCharacters() {
+        char a[] = "abcef".toCharArray();
+        char b[] = "abcef".toCharArray();
+        assertTrue(checkValidityOfText(new String(a), b));
+        assertFalse(checkValidityOfText(new String(a), "ghij".toCharArray()));
+    }
+
```
allanmckenzie: This test is likely to fail randomly. Just because it is a 1/100 probablilyt of being true, doesn't mean that once in every hundred times is true. It could run 1000 times and still not return true (1/10 probability)

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/LocalPartGenerator.java
```diff
@@ -0,0 +1,106 @@
...
+ * The local name part of an email address based on the below rfc <br>
+ * https://tools.ietf.org/html/rfc3696
+ *
+ */
+public class LocalPartGenerator {
+
+    /**
+     * The minimum length of the local part
+     */
+    public static final int MIN_LENGTH = 1;
```
matt-rich: Do these need to be public?

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/LocalPartGeneratorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Test
+    public void shouldCheckAllNonStandardCharacters() {
+        int index = 0;
+        for (int character : new int[] {64, 92, 34, 44, 91, 93}) {
+            assertEquals(character, NONSTANDARD_CHARACTERS[index++]);
+        }
+    }
+
```
allanmckenzie: Not sure if this test is needed

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/EmailAddressGeneratorTest.java
deepakhalale: yeah I agree...the test would still work if all the tests just contain "@" in them. Better to come up with a regex to validate against.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/LocalPartGenerator.java
gogit: yes

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtil.java
```diff
@@ -0,0 +1,121 @@
...
+     * Generate a random string from an array of characters
+     * 
+     * @param validChars to chose from
+     * @param min length of the string
+     * @param max length of the string
+     * @return
+     */
+    public static String generateStringFromCharacters(final Random random, final char[] validChars, final int min,
+                    final int max) {
+        final int size = ThreadLocalRandom.current().nextInt(min, max + 1);
```
deepakhalale: Can we use the provided RANDOM instead of ThreadLocalRandom, as it would be useful to mock for testing.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtil.java
```diff
@@ -0,0 +1,121 @@
...
+     * If the bound is n, returns approximately 1 in n as true <br>
+     * If the bound is 6, returns approximately 1 in 6 as true <br>
+     * If the bound is 2, returns approximately 1 in 2 as true
+     * <br>
+     * 
+     * @param bound
+     * @return flag is true randomly within a bound
+     */
+    public static boolean isRandomlyTrue(final int bound) {
+        return new Random().nextInt(bound) == 0;
```
deepakhalale: Can we use the provided RANDOM instead of creating a new Random, as it would be useful to mock for testing.

gogit: its not related

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtil.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Choose a random position given a length
+     * 
+     * @param number
+     * @return
+     */
+    public static int chooseRandomPosition(final int length) {
+        return new Random().nextInt(length);
```
gogit: its not related

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/GeneratorUtil.java
gogit: GeneratorUtil is a utility class not related to the inheritance tree of generators

gogit: not related to the random parameter passed in

deepakhalale: I don't see how that's not related here.

gogit: as discussed

PR #252, opened 2016-11-03
Refactor Integer Generator

gogit:
Refactor Integer Generator

### DIFF
--- a/test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/IntegerGenerator.java
+++ b/test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/IntegerGenerator.java
@@ -1,15 +1,57 @@
 package uk.gov.justice.services.test.utils.core.random;
 
+import static java.lang.String.format;
+
 public class IntegerGenerator extends Generator<Integer> {
 
-    private final Integer max;
+    /**
+     * The minimum value of the range
+     */
+    private final int min;
 
-    public IntegerGenerator(final Integer max) {
+    /**
+     * The maximum value of the range
+     */
+    private final int max;
+
+    /**
+     * Package Access only
+     * 
+     * @see RandomGenerator
+     */
+    IntegerGenerator() {
+        this.min = Integer.MIN_VALUE;
+        this.max = Integer.MAX_VALUE;
+    }
+
+    /**
+     * Package Access only
+     * 
+     * @param max value excluded
+     * @see RandomGenerator
+     */
+    IntegerGenerator(final int max) {
+        this.min = 0;
+        this.max = max;
+    }
+    /**
+     * Package Access only
+     * 
+     * @param min value included
+     * @param max value excluded
+     * @see RandomGenerator
+     */
+    IntegerGenerator(final int min, final int max) {
+        if (min >= max) {
+            throw new IllegalArgumentException(
+                            format("Min value cannot be greater than or equal to Max value, got Min: %s and Max: %s", min, max));
+        }
+        this.min = min;
         this.max = max;
     }
 
     @Override
     public Integer next() {
-        return RANDOM.nextInt(max);
+        return RANDOM.ints(min, max).limit(1).findFirst().getAsInt();
     }
 }
--- a/test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/IntegerGeneratorTest.java
+++ b/test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/IntegerGeneratorTest.java
@@ -1,0 +1,219 @@
+package uk.gov.justice.services.test.utils.core.random;
+
+import static java.lang.String.format;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class IntegerGeneratorTest {
+
+    @Test
+    public void shouldReturnIntegersInOnesNegativeRange() {
+        shouldTestRangesUsingMinMaxGenerator(-9, 0, 1000);
+    }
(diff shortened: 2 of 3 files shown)


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/IntegerGenerator.java
```diff
@@ -1,15 +1,64 @@
...
+     * Package Access only
+     * 
+     * @param min value included
+     * @param max value excluded
+     * @see RandomGenerator
+     */
+    IntegerGenerator(final int min, final int max) {
+        if (min >= max) {
+            throw new IllegalArgumentException(
+                            format("%s cannot be greater or equal to %s", min, max));
```
deepakhalale: Could have better error message? 
E.g.
Min value cannot be greater than or equal to Max value, got Min: %s and Max: %s

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/IntegerGenerator.java
```diff
@@ -1,15 +1,64 @@
...
         this.max = max;
     }
 
     @Override
     public Integer next() {
-        return RANDOM.nextInt(max);
+        if(min >= 0){
+            return ThreadLocalRandom.current().nextInt(min, max);
+        }
+        long posMin = -(long)min;
```
deepakhalale: Additional variable posMin not required. Same result can be achieved without that.

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/IntegerGenerator.java
```diff
@@ -1,15 +1,64 @@
...
+        long range = (long)max + posMin;
```
deepakhalale: Need final on the variable

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/random/IntegerGenerator.java
```diff
@@ -1,15 +1,64 @@
...
+        }
+        this.min = min;
         this.max = max;
     }
 
     @Override
     public Integer next() {
-        return RANDOM.nextInt(max);
+        if(min >= 0){
+            return ThreadLocalRandom.current().nextInt(min, max);
```
deepakhalale: Any benefit of using ThreadLocalRandom.current().nextInt(min, max) over RANDOM.ints(min, max).limit(1).findFirst().getAsInt(). The reason I am mentioning is we already have RANDOM as part of the parent class and it's easy to mock the RANDOM in testing, if required.

gogit: true its a round about way but has its merits for mocking

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/IntegerGeneratorTest.java
```diff
@@ -0,0 +1,218 @@
...
+    public void shouldReturnIntegersBetweenMinimumInclusiveAndMaximumExcluded() {
+        IntegerGenerator ig = new IntegerGenerator();
+        for (int i = 0; i < 10000; i++) {
+            int k = ig.next();
+            assertTrue(k < Integer.MAX_VALUE);
+            assertTrue(k >= Integer.MIN_VALUE);
+        }
+    }
+       
+    private void shouldTestRangesUsingMinMaxGenerator(final int min, final int max, final int loopCount) {
```
deepakhalale: Same as below

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/IntegerGeneratorTest.java
```diff
@@ -0,0 +1,218 @@
+package uk.gov.justice.services.test.utils.core.random;
+
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class IntegerGeneratorTest {
```
deepakhalale: Not sure if we are adding any value by having lots of test cases which are only testing different ranges? I will leave someone else (Allan or Matt) to decide on this.

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/IntegerGeneratorTest.java
gogit: after we refactor it to make it more readable

test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/IntegerGeneratorTest.java
gogit: if you are generating random numbers, range checking ensures that it works in those ranges and not just in a blanket range.

deepakhalale: Need a test case which tests using Integer.MIN_VALUE and Integer.MAX_VALUE. This is to check that code handles the overflow case.

gogit: Does not apply in this generator

PR #255, opened 2016-11-07
Refactor Long Generator

gogit:
Refactor Long Generator


gogit:
The master code seems to be broken, closing it for the moment, will raise it again after master is fixed

PR #258, opened 2016-11-07
Refactor Random Generator Test

gogit:
Refactor random generator test (Step1) and formatted code to confirm to style

### DIFF
--- a/test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/RandomGeneratorTest.java
+++ b/test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/RandomGeneratorTest.java
@@ -6,6 +6,7 @@ import static java.lang.Boolean.TRUE;
 import static java.math.BigDecimal.ROUND_HALF_EVEN;
 import static java.time.LocalDate.now;
 import static java.util.EnumSet.allOf;
+import static org.junit.Assert.assertNotNull;
 import static uk.gov.justice.services.test.utils.core.helper.TypeCheck.Times.times;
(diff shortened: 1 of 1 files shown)


test-utils/test-utils-core/src/test/java/uk/gov/justice/services/test/utils/core/random/RandomGeneratorTest.java
```diff
@@ -276,8 +277,41 @@ public void shouldAlwaysPickSameElementFromAnEnumWithSingleElement() {
...
-        typeCheck(enumGenerator, s -> enumGenerator.next().compareTo(enumGenerator.next()) == 0).verify(times(NUMBER_OF_TIMES));
+        typeCheck(enumGenerator, s -> enumGenerator.next().compareTo(enumGenerator.next()) == 0)
+                        .verify(times(NUMBER_OF_TIMES));
     }
 
-    enum SingleEnum {SINGLE_VALUE}
-}
\ No newline at end of file
+    @Test
+    public void shouldReturnAnInstanceEmailAddressGenerator() {
```
deepakhalale: Don't see much value in these tests. I think testing EMAIL_ADDRESS that it generates valid email address would naturally test that we get an instance of EMAIL_ADDRESS object!. Same applies for below tests.

gogit: True, they don't add much value like the other tests, after the refactor moved the functionality to each generator. 
They are superficial tests to test the reference exists in the RandomGenerator and has not been removed in a refactor.

deepakhalale:
These tests are fixed in a different PR now. Hence close this down.

PR #266, opened 2016-11-18
add equals and hashcode to FileData

allanmckenzie:


file/file-api/src/main/java/uk/gov/justice/services/file/api/FileData.java
```diff
@@ -20,4 +22,26 @@ public String getFileId() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getFileId(), getFileMimeType());
+    }
+
+    @Override
+    public String toString() {
+        return "FileData{" +
```
matt-rich: Are we mixing single and double quotes because it looks clearer than "'" ?

allanmckenzie: Not sure - I used the default Iintellij create toString() method

allanmckenzie: :-)

PR #267, opened 2016-11-21
Add Streaming support to file service

amitxsethi:
Streaming support to file service

### DIFF
--- a/file/file-alfresco/src/main/java/uk/gov/justice/services/file/alfresco/requester/AlfrescoFileRequester.java
+++ b/file/file-alfresco/src/main/java/uk/gov/justice/services/file/alfresco/requester/AlfrescoFileRequester.java
@@ -11,18 +11,20 @@ import static uk.gov.justice.services.file.alfresco.common.Headers.headersWithUs
 import uk.gov.justice.services.common.configuration.GlobalValue;
 import uk.gov.justice.services.file.alfresco.common.AlfrescoRestClient;
 import uk.gov.justice.services.file.api.FileOperationException;
-import uk.gov.justice.services.file.api.requester.FileRequester;
+import uk.gov.justice.services.file.api.requester.StreamingFileRequester;
 
+import java.io.InputStream;
 import java.util.Optional;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.ws.rs.NotFoundException;
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.StatusType;
 
 @ApplicationScoped
-public class AlfrescoFileRequester implements FileRequester {
+public class AlfrescoFileRequester implements StreamingFileRequester {
     private static final String URL_SEPARATOR = "/";
 
     @Inject
@@ -37,10 +39,9 @@ public class AlfrescoFileRequester implements FileRequester {
     AlfrescoRestClient restClient;
 
     @Override
-    public Optional<byte[]> request(final String fileId, final String fileMimeType, final String fileName, final boolean stream) {
-
+    public Optional<byte[]> request(final String fileId, final String fileMimeType, final String fileName) {
         try {
-            final Response response = restClient.get(alfrescoUriOf(fileId, fileName, stream),
+            final Response response = restClient.get(alfrescoUriOf(fileId, fileName, false),
                     valueOf(fileMimeType), headersWithUserId(alfrescoReadUser));
             final StatusType responseStatus = response.getStatusInfo();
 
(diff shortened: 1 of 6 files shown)


file/file-alfresco/src/test/java/uk/gov/justice/services/file/alfresco/requester/AlfrescoFileRequesterTest.java
```diff
@@ -0,0 +1,128 @@
...
+    @InjectMocks
+    AlfrescoFileRequester alfrescoFileRequester;
+
+    @Test
+    public void shouldRequestAFileAndReceiveByteArrayWhenFileIsFound() {
+        alfrescoFileRequester.alfrescoWorkspacePath = "/service/api/node/content/workspace/SpacesStore/";
+        alfrescoFileRequester.alfrescoReadUser = randomUUID().toString();
+
+        final String fileId = randomUUID().toString();
+        final String fileMimeType = MediaType.TEXT_PLAIN;
```
matt-rich: can static import and below. As values are repeated could make them fields

file/file-api/src/main/java/uk/gov/justice/services/file/api/requester/FileRequester.java
```diff
@@ -7,6 +7,14 @@
...
-    Optional<byte[]> request(final String fileId, final String fileMimeType, final String fileName);
+    /**
+     * Requests a file from the FileService.
+     * @param fileId - the unique id of the file.
+     * @param fileMimeType - mime-type of the file.
+     * @param fileName - name of the file.
+     * @param stream - true to get files in stream mode, else false.
+     * @return content of the file in byte[].
+     */
+    Optional<byte[]> request(final String fileId, final String fileMimeType, final String fileName, final boolean stream);
```
matt-rich: Okay, so as mentioned the other day.. maybe run this by David and see his opinion...

Would it be cleaner to have a FileRequester and a SteamingFileRequester interface (Which extends the FileRequester) ?

It means if we have a file service that can't do streaming (and we know it won't be needed to) we can use the simplified method rather than having to pass false each time.

file/file-alfresco/src/main/java/uk/gov/justice/services/file/alfresco/requester/AlfrescoFileRequester.java
```diff
@@ -51,19 +51,23 @@
...
+    private String alfrescoUriOf(final String fieldId, final String fileName, final boolean stream) {
         final StringBuilder requestBuilder = new StringBuilder();
         requestBuilder.append(alfrescoWorkspacePath);
         requestBuilder.append(fieldId);
         requestBuilder.append(URL_SEPARATOR);
         requestBuilder.append("content");
         requestBuilder.append(URL_SEPARATOR);
         requestBuilder.append(fileName);
+        // a means attach -> a = true means don't stream!
+        if (!stream) {
```
matt-rich: When we send the reply do we not need to do something different as well?

For the request this may be the only modification required but I thought to return it we had to do something differently? maybe not? or maybe it is already implemented (just can't see on Github). Could you just double check please :) - PM me if you need more discussion.

file/file-api/src/main/java/uk/gov/justice/services/file/api/requester/StreamingFileRequester.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.services.file.api.requester;
+
+import java.io.InputStream;
+import java.util.Optional;
+
+public interface StreamingFileRequester extends FileRequester{
+    /**
```
jaceko: not sure if extending the FileRequester interface is a good idea. Maybe we should keep those separate and make Alfresco implement both of them for the time being.

file/file-api/src/main/java/uk/gov/justice/services/file/api/requester/StreamingFileRequester.java
```diff
@@ -0,0 +1,15 @@
...
+     * Requests a file from the FileService as an InputStream.
+     * @param fileId - the unique id of the file.
+     * @param fileMimeType - mime-type of the file.
+     * @param fileName - name of the file.
+     * @return content of the file as InputStream.
+     */
+    Optional<InputStream> requestStreamed(final String fileId, final String fileMimeType, final String fileName);
+}
```
jaceko: maybe just "request" is a better name?

file/file-alfresco/src/main/java/uk/gov/justice/services/file/alfresco/requester/AlfrescoFileRequester.java
```diff
@@ -11,18 +11,20 @@
...
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.ws.rs.NotFoundException;
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.StatusType;
 
 @ApplicationScoped
-public class AlfrescoFileRequester implements FileRequester {
+public class AlfrescoFileRequester implements StreamingFileRequester {
```
jaceko: Maybe just create a new class AlfrescoStreamgFileRequester and keep the interfaces separate? 
@matt-rich what do you think?

file/file-api/src/main/java/uk/gov/justice/services/file/api/requester/StreamingFileRequester.java
amitxsethi: We can't name them request as the FileRequester also has method called request, so the classes implementing both interfaces would have issue

PR #274, opened 2016-11-29
Improve ResponsePayloadMatcher to support string matchers on response payload

deepakhalale:


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/matchers/ResponsePayloadMatcher.java
```diff
@@ -14,24 +18,34 @@
...
         return new ResponsePayloadMatcher();
     }
 
     public ResponsePayloadMatcher isJson(final Matcher<? super ReadContext> matcher) {
-        this.matcher = new IsJson<>(matcher);
+        this.jsonMatcher = Optional.of(new IsJson<>(matcher));
+        return this;
+    }
+
+    public ResponsePayloadMatcher is(final Matcher<String> matcher) {
```
mapingo: renaming method to "that" may sound better

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/http/RestPoller.java
```diff
@@ -63,6 +63,31 @@
...
+ *
+ *      final RequestParams requestParams = requestParams(url, mediaType)
+ *              .withHeader("header-name", "header-value")
+ *              .build();
+ *
+ *      poll(requestParams)
+ *          .until(
+ *              status().is(OK),
+ *              payload()
+ *                  .is(
```
mapingo: I think "that" maybe better here

deepakhalale: Done

PR #297, opened 2017-01-23
refactor and test message consumer client

allanmckenzie:


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/messaging/MessageConsumerClient.java
```diff
@@ -3,46 +3,56 @@
...
 import javax.jms.TextMessage;
 
-import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
-import org.apache.activemq.artemis.jms.client.ActiveMQTopic;
+import com.google.common.annotations.VisibleForTesting;
 
 public class MessageConsumerClient implements AutoCloseable {
 
-    private static final String QUEUE_URI = queueUri();
+    public static final long TIMEOUT_IN_MILLIS = 20_000;
```
matt-rich: these can be private?

test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/messaging/MessageConsumerFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.activemq.artemis.jms.client.ActiveMQTopic;
+
+public class MessageConsumerFactory {
+
+    private Session session;
+    private MessageConsumer messageConsumer;
+    private ActiveMQConnectionFactory connectionFactory;
+    private Connection connection;
+
+    public MessageConsumer createAndStart(
```
matt-rich: IF we're refactoring this I think we should be a bit clearer about Topic/Queues etc.

Can this be used for Queues? - doesn't look like it.. so maybe we need to consider that..

PR #301, opened 2017-01-26
Fix accepted response set for asynchronous httpmethod in raml

mapingo:


generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/helper/Actions.java
```diff
@@ -36,7 +39,8 @@ private Actions() {
     }
 
     public static boolean hasResponseMimeTypes(final Action action) {
```
jaceko: It's not clear to me what this method actually does.
Maybe we should choose a better name for it and/or add a javadoc

mapingo: I agree, renaming and adding docs to this Class

generators/rest-adapter-generator/src/test/java/uk/gov/justice/services/adapters/rest/generator/RestAdapterGenerator_POSTMethodBodyTest.java
```diff
@@ -100,6 +107,41 @@ public void shouldCallInterceptorChainProcessor() throws Exception {
...
+                restRamlWithDefaults().with(
+                        resource("/path")
+                                .with(httpAction(POST)
+                                        .withHttpActionOfDefaultRequestType()
+                                        .withResponsesFrom(responses))
+                ).build(),
+                configurationWithBasePackage(BASE_PACKAGE, outputFolder, emptyMap()));
+
+        Class<?> resourceClass = compiler.compiledClassOf(BASE_PACKAGE, "resource", "DefaultPathResource");
+        Object resourceObject = getInstanceOf(resourceClass);
```
allanmckenzie: final!!!!!!!!!!!

PR #302, opened 2017-01-26
Make TestInjectionPoint resusable accross modules

mapingo:


mapingo:
Adding test, don't merge yet

PR #308, opened 2017-02-01
add logging to rest client

allanmckenzie:


test-utils/test-utils-core/src/main/java/uk/gov/justice/services/test/utils/core/rest/RestClient.java
```diff
@@ -43,11 +60,26 @@ public Response postCommand(final String url, final String contentType, final St
...
+        }
+
+
+        final Response response = clientBuilder().build()
                 .target(url)
                 .request()
                 .headers(headers)
                 .post(entity);
+
+        if (LOGGER.isInfoEnabled()) {
```
matt-rich: could we put these into private methods logResponse and logRequest if these are repeated?

allanmckenzie: A good point - however, by putting the logging into a private method, you loose the correct line number in the logs, which can be kind of useful.

PR #320, opened 2017-02-09
Client generator to ignore non vendor specific media types

jaceko:


generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/client/AbstractClientGenerator.java
```diff
@@ -176,4 +177,9 @@ private FieldSpec loggerConstantField(final String className) {
                 )
                 .build();
     }
+
+    private boolean supportedMimeType(final ActionMimeTypeDefinition mimeType) {
+        final String mimeTypeStr = mimeType.getNameType().getType();
+        return mimeTypeStr.startsWith("application/vnd.") && mimeTypeStr.endsWith("+json");
```
matt-rich: There might be Names constants that define these values (possibly...)

PR #326, opened 2017-02-17
Event stream and snapshots refactoring to cleanup and remove code duplication

jaceko:


aggregate-snapshot/aggregate-snapshot-service/pom.xml
```diff
@@ -118,5 +118,15 @@
             <artifactId>javapoet</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>uk.gov.justice.services</groupId>
+            <artifactId>test-utils-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
matt-rich: test scope dependency?

PR #329, opened 2017-02-17
Fix to generate random ZonedDateTime in different time zones and provide option to generate in UTC specific timezone

deepakhalale:


purple52:
Do we want it to always generate a UTC date? Maybe that should be a specialised form of the generator, or an option when creating it. I would expect a generate random zoned date generator to produce randomised timezones too.

deepakhalale:
Sure will look to fix it. Thanks.

deepakhalale:
Fixed based on review comments.

PR #337, opened 2017-02-21
Make ResponseStrategy open for extension

mapingo:


generators/generators-test-utils/src/main/java/uk/gov/justice/services/generators/test/utils/reflection/ReflectionUtil.java
```diff
@@ -42,13 +42,22 @@ public static void setField(final Object object, final String fieldName, final O
...
         assertTrue(field.isPresent());
         return field.get();
     }
 
     /**
+     * Checks for a field in the given class by reflection
+     *
+     * @return - true if found otherwise false
+     */
+    public static Boolean fieldIsPresent(final Class<?> clazz, final String fieldName) {
```
purple52: Should this return a primitive boolean to avoid nulls?

PR #339, opened 2017-02-28
(1.4) Local jndi value falls back to global value

jaceko:


common/src/main/java/uk/gov/justice/services/common/configuration/GlobalValueProducer.java
```diff
@@ -24,7 +25,7 @@ public String produceValue(final InjectionPoint ip) throws NamingException {
     }
 
     @Override
-    protected String jndiNameFrom(final CommonValueAnnotationDef annotation) {
-        return format("java:global/%s", annotation.key());
+    protected String[] jndiNamesFrom(final CommonValueAnnotationDef annotation) {
+        return new String[]{format("java:global/%s", annotation.key())};
```
matt-rich: possibly make the global a constant

common/src/main/java/uk/gov/justice/services/common/configuration/AbstractValueProducer.java
```diff
@@ -14,15 +15,19 @@ protected AbstractValueProducer() throws NamingException {
...
 
     protected String jndiValueFor(final CommonValueAnnotationDef annotation) throws NamingException {
-        try {
-            return (String) initialContext.lookup(jndiNameFrom(annotation));
-        } catch (NameNotFoundException e) {
-            if (CommonValueAnnotationDef.NULL_DEFAULT.equals(annotation.defaultValue())) {
-                throw new MissingPropertyException(format("Missing property: %s", annotation.key()));
+        for (String jndiName : jndiNamesFrom(annotation)) {
+            try {
+                return (String) initialContext.lookup(jndiName);
```
matt-rich: if we have 2 loops, won't the return exit out of the loop?
Is this purely going to work because the order of the array (app then global) ?

jaceko: yes

matt-rich: Can we add a javadoc to explain that.

jaceko: done

common/src/main/java/uk/gov/justice/services/common/configuration/GlobalValueProducer.java
jaceko: done

common/src/main/java/uk/gov/justice/services/common/configuration/AbstractValueProducer.java
matt-rich: I like that solution 👍

PR #352, opened 2017-03-16
Add MDB pool configuration

jaceko:


example-context/example-service/example-event/example-event-listener/pom.xml
```diff
@@ -23,7 +23,11 @@
             <artifactId>javaee-api</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
```
matt-rich: Could we add this  dependency (as it is provided) to the component modules?, so if anybody does want to start using the pool they don't need to add it.

Also presumably this works for Event-Processor, Command Handler MDBs too?

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
```diff
@@ -111,21 +118,32 @@ TypeSpec generatedCodeFor(final Resource resource, final BaseUri baseUri, final
...
-                    .addMember(DEFAULT_ANNOTATION_PARAMETER, "$T.class", JmsLoggerMetadataInterceptor.class)
-                    .addMember(DEFAULT_ANNOTATION_PARAMETER, "$T.class", JsonSchemaValidationInterceptor.class)
-                    .build());
-        }
+    private String poolNameFrom(final MessagingResourceUri resourceUri, final String component) {
+        return format("%s-%s-pool", resourceUri.hyphenated(), component.toLowerCase().replace("_", "-"));
+    }
+
+    private boolean shouldAddCustomPoolConfiguration(final GeneratorConfig configuration) {
+        return configuration.getGeneratorProperties().containsKey(CUSTOM_MDB_POOL_CONFIG_PROPERTY)
```
matt-rich: Does the GET throw an exception or return null?

Generally speaking it is better to do the .equals() on the constant as it will never be a null pointer. Can this be simplified to "TRUE".equalsIgnoreCase(configuration.getGerenatorPropertions().get(CUSTOM_...) ?

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
```diff
@@ -111,21 +118,32 @@ TypeSpec generatedCodeFor(final Resource resource, final BaseUri baseUri, final
...
+                        .build());
+            }
 
             return typeSpecBuilder;
         }
 
         throw new IllegalStateException(format("JMS Endpoint generation is unsupported for framework component type %s", component));
     }
 
-    private void addLoggerAndSchemaInterceptorsIfNotGeneralJsonMimeType(final Resource resource, final TypeSpec.Builder typeSpecBuilder) {
```
matt-rich: I actually prefer the wrapper name than the inline.. but no big deal. It hides some of the details from the invoking method.

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
jaceko: if a map containsKey (which I'm checking first) then it's safe to do a get on this map to fetch value of this key

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
jaceko: I just felt that the method name was a bit long and complex and that the code gained on readability when actually inlining the if statement.

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
matt-rich: But if we don't do the contains what happens?
If it just returns null then we can simplify the expression, as false or null.. neither equals TRUE so we don't need the defensive check.

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
matt-rich: Fair enough, personal taste I guess. The discussion before was if you are maintaining the code and didn't know poet etc. so well then it is clearer in a private method. Agreed the name wasn't so good :)

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/MessageListenerCodeGenerator.java
jaceko: OK, changed that.

PR #360, opened 2017-03-27
Avoid race condition in event stream append

### LINKED ISSUE #356: Race condition in EventStream append
When appending an event with no version, the event stream asks the database for the next version number of that stream. But if someone else has appended an event in the tie between the stream being read and the max version being checked, optimistic locking doesn't kick in because we append our event after the other persons event without complaining.

We need to attempt the event append based on the highest version in the stream we originally read.

purple52:
Fixes #356 by using the maximum version number in the events that have been read instead of getting the version number from the database


event-sourcing/event-source/src/main/java/uk/gov/justice/services/eventsourcing/source/core/EnvelopeEventStream.java
```diff
@@ -49,11 +54,24 @@ public long appendAfter(final Stream<JsonEnvelope> events, final long version) t
 
     @Override
     public long getCurrentVersion() {
```
jaceko: shouldn't that be synchronised too?

PR #361, opened 2017-03-29
Add interceptor chain provider

mapingo:


core/src/main/java/uk/gov/justice/services/core/interceptor/DefaultInterceptorChainProvider.java
```diff
@@ -0,0 +1,26 @@
...
+import uk.gov.justice.services.core.metrics.TotalActionMetricsInterceptor;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+
+
+public abstract class DefaultInterceptorChainProvider implements InterceptorChainProvider {
```
jaceko: I would rename it to something like BaseInterceptorChainProvider.
Classes prefixed with "Default" are usually non-abstract default implementation.

PR #363, opened 2017-03-30
Framework api for generators

mapingo:


framework-api/framework-api-core/src/main/java/uk/gov/justice/services/core/interceptor/ContextPayload.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.justice.services.core.interceptor;
+
+import uk.gov.justice.services.messaging.JsonEnvelope;
+
+import java.util.Map;
+import java.util.Optional;
+
+public interface ContextPayload {
```
jaceko: Could use some javadocs explaining what is the purpose of this interface

PR #377, opened 2017-04-10
Move metrics interceptor into metrics module

jaceko:


metrics/metrics-interceptor/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>uk.gov.justice.services</groupId>
```
matt-rich: if we have test utils do we still need unit and mockito?

PR #380, opened 2017-04-13
Add lint check to example app & refactor pom configs

skiddykong:


example-context/example-service/example-event/example-event-processor/pom.xml
```diff
@@ -35,74 +39,4 @@
...
-                            <goal>generate</goal>
-                        </goals>
-                        <phase>generate-sources</phase>
-                        <configuration>
-                            <generatorName>uk.gov.justice.raml.jms.core.JmsEndpointGenerator</generatorName>
-                            <sourceDirectory>${basedir}/src/raml</sourceDirectory>
-                            <outputDirectory>${project.build.directory}/generated-sources</outputDirectory>
-                            <basePackageName>uk.gov.justice.api</basePackageName>
-                            <includes>
-                                <include>**/*cakeshop-event-processor.messaging.raml</include>
```
matt-rich: Is there anything special about these paths/values that mean they still need to be implemented in the per pom level (just this bit, not the entire config)?

example-context/example-service/example-event/example-event-processor/pom.xml
```diff
@@ -35,74 +39,4 @@
...
-                            <goal>generate</goal>
-                        </goals>
-                        <phase>generate-sources</phase>
-                        <configuration>
-                            <generatorName>uk.gov.justice.services.clients.messaging.generator.MessagingClientGenerator</generatorName>
-                            <sourceDirectory>${basedir}/src/raml</sourceDirectory>
-                            <outputDirectory>${project.build.directory}/generated-sources</outputDirectory>
-                            <basePackageName>uk.gov.justice.api</basePackageName>
-                            <includes>
-                                <include>**/*public-event.messaging.raml</include>
```
matt-rich: Same as above, are we losing anything as we are no longer including this specific path/value ?

skiddykong: I think it is the same as other configurations, the public-event raml will be included in the normal *.messaging.raml pick up

skiddykong: as above

example-context/example-service/example-event/example-event-processor/pom.xml
skiddykong: I don’t see any issues as the public.events.messaging.raml is picked up by *.messaging.raml includes in the parent pom

PR #387, opened 2017-04-28
DSL for Domain Testing

khanhassan:


test-utils/test-utils-domain/src/test/java/uk/gov/justice/services/example/cakeshop/domain/Ingredient.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.services.example.cakeshop.domain;
```
purple52: I think its confusing to have copies of the example context domain here. We should create specific test cases for the Cucumber tests instead.

test-utils/test-utils-domain/src/main/java/uk/gov/justice/services/domain/main/GenericStepDefs.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @When("(.*) to a (.*) using (.*)")
+    public void call_method_with_params(final String methodName, final String aggregate, final String fileName)
+            throws Exception {
+        createAggregate(aggregate);
+        ObjectMapper mapper = new ObjectMapper();//normal mapper to get the map value in correct order
+        Map argumentsMap = mapper.convertValue(mapper.readTree(json(fileName)), Map.class);
+        mapper = new ObjectMapperProducer().objectMapper();
+        List valuesList = new ArrayList(argumentsMap.values());
+        checkIfUUID(valuesList);
```
purple52: This needs to work for all parameter types by using Jackson to convert into the target parameter type for each parameter.

purple52:
Code moved to [domain-test-dsl](https://github.com/CJSCommonPlatform/domain-test-dsl).

PR #389, opened 2017-05-03
Manage direct clients discovery

jaceko:
Discover direct clients only if matching adapter exists; Discover direct clients before non-direct ones.


core/src/main/java/uk/gov/justice/services/core/extension/AnnotationScanner.java
```diff
@@ -45,22 +51,37 @@
 
     @SuppressWarnings("unused")
     void afterDeploymentValidation(@Observes final AfterDeploymentValidation event, final BeanManager beanManager) {
-        beanManager.getBeans(Object.class, AnyLiteral.create())
-                .forEach(this::processBean);
+        final Set<Bean<?>> directAdapters = beanManager.getBeans(SynchronousDirectAdapter.class);
```
purple52: Can we leave this class as-is and create a new Extension that lives in the direct client module that just scans for direct adapters?

PR #398, opened 2017-05-31
Add local dependency install script and documentation

mapingo:
How to build and deploy Microservice Framework locally without access to build repository.  Includes useful script for building and deploying dependencies into the local m2 cache.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+
+# Script to checkout and install locally all Microservice Framework dependencies
+
+#Build versions
+# maven-super-pom
+maven_super_pom_versions=("release-1.0.0")
+
+# maven-parent-pom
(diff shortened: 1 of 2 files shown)


install-dependencies.sh
```diff
@@ -0,0 +1,92 @@
...
+cd ../file-service
+installVersions ${file_service_versions[@]}
+
+cd ../json-schema
+installVersions ${json_schema_versions[@]}
+
+cd ../wildfly-maven-plugin
+installVersions ${wildfly_maven_plugin_versions[@]}
+
+cd ../microservice-framework
```
purple52: Should this last one be an underscore instead of a dash?

mapingo: Yes, I missed that, corrected.

mapingo:
I am not sure what Coveralls is on about here?

purple52:
Yeah, no idea why Coveralls is upset. Let's see what happens when its merged. Thanks!

PR #399, opened 2017-05-31
Fix JavaDocs errors and warnings

mapingo:
Remove all Java Doc errors and warnings, to allow mvn javadoc:javadoc to compile Java Docs without failures.


purple52:
Thanks, Shaun!

PR #410, opened 2017-06-16
Framework API tweaks

purple52:


domain/src/main/java/uk/gov/justice/domain/aggregate/condition/Precondition.java
```diff
@@ -4,7 +4,11 @@
 
 /**
  * Representation of a precondition that can be asserted, and throw an exception if it isn't met.
+ *
+ * NB: This class is not deprecated to discourage use within aggregates. It if finds a use elsewhere
```
matt-rich: not=now deprecated typo? Normally I wouldn't bother (honest ;) ) but it kind of changes the meaning significantly...

purple52: Ha! Well spotted.

PR #419, opened 2017-06-22
Allow any base URI for messaging client class names

purple52:


generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/helper/MessagingClientBaseUri.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class MessagingClientBaseUri {
+
+    private static final Pattern MESSAGING_BASE_URI_PATTERN = Pattern.compile("message://(.+)");
+    private final String uriExcludingProtocol;
+
+    public MessagingClientBaseUri(final String uriString) {
+        final Matcher m = matcherOf(uriString);
+        m.find();
+        this.uriExcludingProtocol = m.group(1);
```
matt-rich: if this is invalid does this present a nice error message?

generators/generators-commons/src/test/java/uk/gov/justice/services/generators/commons/helper/MessagingAdapterBaseUriTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    public void shouldReturnComponent() {
+        assertThat(new MessagingAdapterBaseUri("message://event/listener/message/service1").component(), is(EVENT_LISTENER));
+        assertThat(new MessagingAdapterBaseUri("message://event/processor/message/service2").component(), is(EVENT_PROCESSOR));
+    }
+
+    @Test
+    public void shouldReturnClassName() {
+        assertThat(new MessagingAdapterBaseUri("message://event/listener/message/system").toClassName(), is("SystemEventListener"));
+    }
```
matt-rich: No negative test of an invalid uri?

generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/helper/MessagingClientBaseUri.java
purple52: No it doesn't, as per the original. We need to refactor this completely to avoid making assumptions about the base URI at all.

PR #434, opened 2017-08-15
Add event stream repository & rename eventLog to event

jaceko:


event-sourcing/event-repository/event-repository-jdbc/src/main/java/uk/gov/justice/services/eventsourcing/repository/jdbc/event/EventJdbcRepository.java
```diff
@@ -23,14 +24,16 @@
 /**
  * JDBC based repository for event log records.
  */
-public class EventLogJdbcRepository extends AbstractJdbcRepository<EventLog> {
+@ApplicationScoped
+public class EventJdbcRepository extends AbstractJdbcRepository<Event> {
 
+    private static final String FAILED_TO_READ_STREAM = "Failed to read stream {}";
     //TODO @mrich 2016-09-29 - Re-enable the date_created field once liquibase scripts are configured
```
matt-rich: I think this commented out line should be deleted now, we added the date_created a while ago... (I know it's not your change... but if you're editing :)

PR #449, opened 2017-08-22
remove event payload obj

MohamedFarouk-HMCTS:


event-sourcing/event-source-api-integration-test/src/test/java/uk/gov/justice/services/eventsourcing/source/api/EventsFeedIT.java
```diff
@@ -163,20 +163,17 @@ public void shouldReturnFirstPageOfFeed() throws Exception {
                 .assertThat("$.data[0].streamId", is(streamId.toString()))
                 .assertThat("$.data[0].name", containsString(NAME))
                 .assertThat("$.data[0].sequenceId", is(1))
-                .assertThat("$.data[0].payload.streamId", is(streamId.toString()))
-                .assertThat("$.data[0].payload.payloadContent", containsString(PAYLOAD_JSON))
+                .assertThat("$.data[0].payload", containsString(PAYLOAD_JSON))
```
jaceko: as I said earlier this is a weak assertion. the test will still pass if the payload is 'abcbsbsbbs{\"field\": \"Value\"}'
better to use json matcher to match payload content
it's also better to setup event payload in the test amd maybe diffrenet payload in each event

PR #474, opened 2017-10-30
Remove unused dependencies from test-utils-core

mapingo:
Remove OWASP dependency vulnerability hamcrest-date and unused hamcrest-library


mapingo:
Not to be merged into release-2.3.x

PR #488, opened 2017-11-08
Add POJO support to HandlerMethods - HandlerMethods

skiddykong:


core/src/main/java/uk/gov/justice/services/core/handler/HandlerMethod.java
```diff
@@ -78,7 +78,16 @@ private static boolean isVoid(final Class<?> clazz) {
     }
 
     private static boolean isEnvelope(final Class<?> clazz) {
-        return JsonEnvelope.class.equals(clazz);
```
crossing: Is it possible to use `Class.isAssignableFrom` ?

skiddykong: No such method sadly according to my IDE

purple52: Its an instance method; JsonEnvelope.class.isAssignableFrom(clazz) should work.

core/src/main/java/uk/gov/justice/services/core/handler/HandlerMethodInvoker.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class HandlerMethodInvoker {
+
+    public Object invoke(final Object handlerInstance,
+                         final Method handlerMethod,
+                         final JsonEnvelope jsonEnvelope)
+            throws InvocationTargetException, IllegalAccessException, ClassNotFoundException, IOException {
+
+        final Class<?> parameterClass = handlerMethod.getParameterTypes()[0];
+        if(parameterClass == JsonEnvelope.class) {
```
mapingo: Instead of "==" maybe JsonEnvelope.class.equals(parameterClass) would be better

allanmckenzie: As long as you are in the same virtual machine then equals(...) should be equivalent to '=='

core/src/main/java/uk/gov/justice/services/core/handler/HandlerMethodInvoker.java
```diff
@@ -0,0 +1,46 @@
...
+            return handlerMethod.invoke(handlerInstance, jsonEnvelope);
+        }
+        final Type[] genericParameterTypes = handlerMethod.getGenericParameterTypes();
+        final Type[] parameters = ((ParameterizedType)genericParameterTypes[0]).getActualTypeArguments();
+
+        final Class<?> envelopeType = forName(parameters[0].getTypeName());
+
+        final JsonObject jsonObject = jsonEnvelope.payloadAsJsonObject();
+
+        final ObjectMapper objectMapper = new ObjectMapperProducer().objectMapper();
```
mapingo: It would be good if the ObjectMapper was constructor injected.  This may also lead to the HandlerMethodInvoker also needing to be constructor injected.  Might worth another story for this refactoring.

core/src/main/java/uk/gov/justice/services/core/handler/HandlerMethodInvoker.java
mapingo: Yes you are right...I was thinking of something else...scratch my comment

PR #492, opened 2017-11-21
Add pojo envelope support to Senders Requesters

skiddykong:


core/src/main/java/uk/gov/justice/services/core/dispatcher/DispatcherDelegate.java
```diff
@@ -39,6 +58,16 @@ public void send(final JsonEnvelope envelope) {
         dispatcher.dispatch(envelope);
     }
 
+    @Override
+    public <T> void send(final Envelope<T> envelope) {
```
purple52: This can be a wildcard ? instead of a T

framework-api/framework-api-core/src/main/java/uk/gov/justice/services/core/sender/Sender.java
```diff
@@ -14,6 +15,13 @@
      */
     void send(final JsonEnvelope envelope);
 
+    /**
+     * Sends envelope to the next component.  The correct sender is injected by the framework.
+     *
+     * @param envelope with payload T that needs to be sent.
+     */
+    <T> void send(final Envelope<T> envelope);
```
purple52: Can't this just replace the method above?

PR #505, opened 2017-12-07
Event filter fix

MohamedFarouk-HMCTS:
This is to fix the issue with no schema validations of command handler due to single war class loading EventFilter of Event Listener for handlers which then skips validation due to no match on the names

### DIFF
--- a/generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/JmsEndpointGenerator.java
+++ b/generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/JmsEndpointGenerator.java
@@ -1,6 +1,8 @@
 package uk.gov.justice.raml.jms.core;
 
 import static org.raml.model.ActionType.POST;
+import static uk.gov.justice.raml.jms.core.JmsEndPointGeneratorUtil.shouldGenerateEventFilter;
+import static uk.gov.justice.raml.jms.core.JmsEndPointGeneratorUtil.shouldListenToAllMessages;
 import static uk.gov.justice.raml.jms.core.MediaTypesUtil.containsGeneralJsonMimeType;
 import static uk.gov.justice.services.core.annotation.Component.EVENT_LISTENER;
 import static uk.gov.justice.services.generators.commons.helper.GeneratedClassWriter.writeClass;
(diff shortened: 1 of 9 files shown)


generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/JmsEndpointGenerator.java
```diff
@@ -79,11 +79,11 @@ public void run(final Raml raml, final GeneratorConfig configuration) {
     therefore we need all messages with consecutive numbers there. Messages need to be in correct order in order to update the view correctly.
     */
 
-    private boolean shouldGenerateEventFilter(final Resource resource, final MessagingAdapterBaseUri baseUri) {
+    static boolean shouldGenerateEventFilter(final Resource resource, final MessagingAdapterBaseUri baseUri) {
```
mapingo: If we are making these two methods static and accessible from another class, then shouldn't they be in another Class?

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/JmsEndpointGenerator.java
```diff
@@ -79,11 +79,11 @@ public void run(final Raml raml, final GeneratorConfig configuration) {
...
         return EVENT_LISTENER.equals(baseUri.component()) && !containsGeneralJsonMimeType(resource.getActions());
     }
 
-    private boolean shouldListenToAllMessages(final Resource resource, final MessagingAdapterBaseUri baseUri) {
+    static boolean shouldListenToAllMessages(final Resource resource, final MessagingAdapterBaseUri baseUri) {
```
mapingo: As above

MohamedFarouk-HMCTS: Moved to a separate class JmsEndPointGeneratorUtil

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/JmsEndpointGenerator.java
MohamedFarouk-HMCTS: Moved to a separate class JmsEndPointGeneratorUtil

PR #509, opened 2017-12-13
Add schema catalog generation and validation

mapingo:


core/src/main/java/uk/gov/justice/services/core/json/SchemaCatalogAwareJsonSchemaValidator.java
```diff
@@ -0,0 +1,67 @@
...
+    public void validate(final String envelopeJson, final MediaType mediaType) {
+
+        final Optional<Schema> schema = getSchema(mediaType);
+        if (schema.isPresent()) {
+            logger.trace("Performing schema validation with catalog schema for: {}", mediaType);
+            final JSONObject payload = payloadExtractor.extractPayloadFrom(envelopeJson);
+            schema.get().validate(payload);
+        } else {
+            logger.trace("Falling back to file base schema lookup, no catalog schema found for: {}", mediaType);
+            doValidationWithoutMappedSchema(envelopeJson, mediaType);
```
allanmckenzie: I like this log message. Could be useful. Maybe promote it to info?

core/src/main/java/uk/gov/justice/services/core/mapping/SchemaIdMappingObserver.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.slf4j.Logger;
+
+/**
+ * Collects all {@link MediaTypeToSchemaIdMapper} beans that are annotated with
+ * {@link SchemaIdMapper} after deployment, and allows access to list.
+ */
+public class SchemaIdMappingObserver implements Extension {
+
+    private static final Logger LOGGER = getLogger(SchemaIdMappingObserver.class);
```
allanmckenzie: I'm guessing that you're using a static LOGGER rather than an injected one, it that this class is not managed by CDI. Otherwise the logger should be injected

generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/mapping/MediaTypeToSchemaIdGenerator.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Generates MediaType to schema id mapping class for a given RAML object
+ */
+public class MediaTypeToSchemaIdGenerator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MediaTypeToSchemaIdGenerator.class);
+    private final SchemaMappingClassNameGenerator schemaMappingClassNameGenerator = new SchemaMappingClassNameGenerator();
+    private final MediaTypeToSchemaIdParser mediaTypeToSchemaIdParser = new MediaTypeToSchemaIdParser(new SchemaIdParser());
+    private final MediaTypeToSchemaIdMapperClassBuilder mediaTypeToSchemaIdMapperClassBuilder = new MediaTypeToSchemaIdMapperClassBuilder(schemaMappingClassNameGenerator);
+
```
allanmckenzie: Can we use injection here - rather than instantiating everything in the class? Or is this another example of were we can't

generators/generators-commons/src/test/java/uk/gov/justice/services/generators/commons/mapping/SchemaMappingClassNameGeneratorTest.java
```diff
@@ -0,0 +1,42 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class SchemaMappingClassNameGeneratorTest {
+
+    @InjectMocks
+    private SchemaMappingClassNameGenerator schemaMappingClassNameGenerator;
+
+    @Test
+    public void shouldCreateTheSchemaMappingClassNameFromTheContextNameAndItsInterfaceName() throws Exception {
+
+        final String baseUri = "http://localhost:8080/people-command-api/command/api/rest/people";
```
allanmckenzie: Maybe anonymise this url.

core/src/main/java/uk/gov/justice/services/core/mapping/SchemaIdMappingObserver.java
mapingo: If the Logger is injected this does not work with the OpenEJB tests...maybe there is a work around...but keeping as it is for now.

generators/generators-commons/src/main/java/uk/gov/justice/services/generators/commons/mapping/MediaTypeToSchemaIdGenerator.java
mapingo: This is created from the generator plugin...no access to CDI from here

PR #510, opened 2017-12-13
Resolve Single War Schema Conflicts

matt-rich:


event-sourcing/event-repository/event-repository-jdbc/pom.xml
```diff
@@ -31,6 +31,11 @@
             <groupId>uk.gov.justice.utils</groupId>
             <artifactId>utilities-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>uk.gov.justice.utils</groupId>
+            <artifactId>test-utils-core</artifactId>
+            <version>${test-utils.version}</version>
+        </dependency>
```
matt-rich: This should probably be test scope

PR #522, opened 2018-01-17
Remove components and add interceptors modules

mapingo:


framework-bom/pom.xml
```diff
@@ -83,47 +83,12 @@
 
             <dependency>
                 <groupId>uk.gov.justice.services</groupId>
-                <artifactId>command-api</artifactId>
+                <artifactId>command-handler-interceptors</artifactId>
```
matt-rich: Might be useful to group the two interceptor dependecies next to each other?

PR #540, opened 2018-01-30
SPI for InterceptorContext.interceptorContextWithInput()

artur-synowiec:


core/src/main/java/uk/gov/justice/services/core/interceptor/DefaultInterceptorContext.java
```diff
@@ -33,7 +33,9 @@ private DefaultInterceptorContext(final ContextPayload input, final ContextPaylo
      *
      * @param input the input JsonEnvelope
      * @return the new InterceptorContext
+     * @deprecated Use InterceptorContext.interceptorContextWithInput() provided by DefaultInterceptorContextProvider
      */
+    @Deprecated
     public static InterceptorContext interceptorContextWithInput(final JsonEnvelope input) {
         return new DefaultInterceptorContext(contextPayloadWith(input), contextPayloadWithNoEnvelope());
```
mapingo: Can this be changed to use Interceptor.interceptorContextWithInput(input) as the implementation?

PR #541, opened 2018-01-31
Add support for 404 results for queries for pojo envelopes

skiddykong:


core/src/main/java/uk/gov/justice/services/core/dispatcher/EnvelopePayloadTypeConverter.java
```diff
@@ -38,6 +38,10 @@ public EnvelopePayloadTypeConverter() {
             return null;
         }
 
+        if(envelope.payload() == null){
```
purple52: Needs a test case

generators/rest-adapter-core/src/main/java/uk/gov/justice/services/adapter/rest/processor/response/ResponseStrategyHelper.java
```diff
@@ -41,7 +41,7 @@ Response responseFor(final String action, final Optional<JsonEnvelope> result, f
         if (result.isPresent()) {
             final JsonEnvelope outputEnvelope = result.get();
 
-            if (outputEnvelope.payload() == NULL) {
+            if (outputEnvelope.payload() == NULL || outputEnvelope.payload() == null) {
```
purple52: Check should probably be the other way around - check for null first then NULL. Also needs a test case!

core/src/main/java/uk/gov/justice/services/core/dispatcher/EnvelopePayloadTypeConverter.java
skiddykong: Done.

PR #543, opened 2018-02-06
Update dbcleaner test util to clean the event_stream table

matt-rich:


test-utils/test-utils-persistence/src/main/java/uk/gov/justice/services/test/utils/persistence/DatabaseCleaner.java
```diff
@@ -77,6 +77,7 @@ public void cleanEventLogTable(final String contextName) {
         try (final Connection connection = testJdbcConnectionProvider.getEventStoreConnection(contextName)) {
 
             cleanTable("event_log", connection);
+            cleanTable("event_stream", connection);
```
MohamedFarouk-HMCTS: cleanTable("event_stream", connection); 
would prefer a separate method to clean event stream as it deviates from the pattern of other table deletes also the method name is called clearEventLog and it deletes EventStream table

matt-rich: Updated the method name, deprecated the original method from this release.

We shouldn't allow people to delete 'any' table from the event_store as the event_store is a framework controlled module and therefore we should know which tables should be deleted. The viewstore is special as the developer decides which tables to create (different for each projection/service)

PR #548, opened 2018-02-23
Master branch should be consistent so that projects can build without…

ashtonhogan:
… manual intervention


purple52:
Thanks for the pull request, but we don't want master on this project to be building against a snapshot of another project. Apart from anything else, we don't keep snapshot builds in Bintray to Travis can't build the project if it depends on snapshots.

PR #550, opened 2018-02-28
Support pojo envelopes in enveloper

BenNzewi:


core/src/main/java/uk/gov/justice/services/core/enveloper/spi/DefaultEnveloperProvider.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.services.core.enveloper.spi;
+
+import uk.gov.justice.services.core.enveloper.DefaultEnveloper;
+import uk.gov.justice.services.core.enveloper.Enveloper;
+import uk.gov.justice.services.messaging.Envelope;
+import uk.gov.justice.services.messaging.JsonEnvelope;
+
+import java.util.function.Function;
+
+public class DefaultEnveloperProvider implements EnveloperProvider {
```
purple52: This class needs to bridge the gap between the static, SPI world and the CDI managed bean world. The DefaultEnveloper should be a managed bean, and this non-managed class should get a reference to that bean and use it to call non-static methods. Currently, you are using static field in the DefaultEnveloper to trick CDI into injecting beans into its instance of the DefaultEnveleper and using the static fields to pass those beans to the instance that SPI has created, which is all a bit of a nasty workaround.

core/src/main/java/uk/gov/justice/services/core/enveloper/DefaultEnveloper.java
```diff
@@ -101,12 +113,40 @@ private Metadata buildMetaData(final Metadata metadata, final String name) {
...
+        @Override
+        public EnveloperBuilder withName(final String name) {
+            this.setName(name);
+            return this;
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public Envelope withMetadataFrom(final Envelope envelope) {
+            return new DefaultEnvelope(buildMetaData(envelope.metadata(), this.name, clock), this.payload);
```
purple52: Should this use envelopeFrom rather than the constructor?

example-context/example-service/example-command/example-command-handler/src/main/java/uk/gov/justice/services/example/cakeshop/command/handler/MakeCakeCommandHandler.java
```diff
@@ -51,9 +47,7 @@ public void makeCake(final JsonEnvelope command) throws EventStreamException {
 
         eventStream.append(
                 recipe.makeCake(cakeId)
-                        .map(enveloper.withMetadataFrom(command)),
-                Tolerance.CONSECUTIVE);
```
purple52: Was the Tolerance.CONSECUTIVE dropped deliberately?

example-context/example-service/example-command/example-command-handler/src/main/java/uk/gov/justice/services/example/cakeshop/command/handler/MakeCakeCommandHandler.java
```diff
@@ -51,9 +47,7 @@ public void makeCake(final JsonEnvelope command) throws EventStreamException {
...
-
+                        .map(Enveloper.toEnvelopeWithMetadataFrom(command)));
```
purple52: method should be static imported

messaging/messaging-core/src/main/java/uk/gov/justice/services/messaging/spi/DefaultEnvelope.java
```diff
@@ -6,13 +6,13 @@
...
  */
-public class DefaultEnvelope <T> implements Envelope<T> {
+public class DefaultEnvelope<T> implements Envelope<T> {
 
     private final Metadata metadata;
 
     private final T payload;
 
-    DefaultEnvelope(final Metadata metadata, final T payload) {
+    public DefaultEnvelope(final Metadata metadata, final T payload) {
```
purple52: Do we have to make this public? We don't want people creating these directly.

PR #551, opened 2018-03-01
Add message adapter generator support for multiple event listeners

mapingo:


generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/core/ComponentDestinationType.java
```diff
@@ -40,15 +40,39 @@ public ComponentDestinationType() {
...
             return components.get(component);
         }
 
         throw new IllegalArgumentException("No input destination type defined for service component of type " + component);
     }
 
-
     public boolean isSupported(final String component) {
-        return components.containsKey(component);
+        return component.contains(COMMAND_API)
```
matt-rich: Isn''t this just always going to return true now?

mapingo: not quite...some components are not on the list of supported components
        COMMAND_API
        COMMAND_CONTROLLER
        COMMAND_HANDLER
        EVENT_PROCESSOR
        EVENT_LISTENER
Are the only components supported plus custom versions of theses.

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/interceptor/EventListenerInterceptorChainProviderCodeGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+ *          @literal @Override
+ *          public List<InterceptorChainEntry> interceptorChainTypes() {
+ *              return interceptorChainEntries;
+ *          }
+ *     }
+ *
+ * }
+ * </pre>
+ */
+public class EventListenerInterceptorChainProviderCodeGenerator {
```
matt-rich: Does this still work with people wanting to create / append their own providers?

Such as the authorisation interceptor and things?

interceptors/event-listener-interceptors/src/main/java/uk/gov/justice/services/components/event/listener/interceptors/EventFilterInterceptor.java
```diff
@@ -15,8 +15,10 @@
 
     @Override
```
matt-rich: Isn't this a generated class now, with the specific EventFilter class not injected?

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/interceptor/EventListenerInterceptorChainProviderCodeGenerator.java
mapingo: People are still able to add into the Interceptor Chain Provider by implementing another InterceptorChainEntryProvider....both providers are loaded from the classpath and combined at start up...they will not be able to remove from the chain though

interceptors/event-listener-interceptors/src/main/java/uk/gov/justice/services/components/event/listener/interceptors/EventFilterInterceptor.java
mapingo: This can be removed...but probably should be deprecated first...I did deprecate in another pull request to support multiple topics

matt-rich: Sounds good

generators/messaging-adapter-generator/src/main/java/uk/gov/justice/raml/jms/interceptor/EventListenerInterceptorChainProviderCodeGenerator.java
matt-rich: ok cool, I assumed so but just wanted to check.

PR #559, opened 2018-03-20
Update message generator

shivkumaryadav:


generators/messaging-adapter-generator/src/main/java/uk/gov/justice/subscription/jms/core/ClassNameFactory.java
```diff
@@ -0,0 +1,64 @@
...
+        final String simpleName =
+                        buildJavaFriendlyName(contextName) +
+                        componentNameToClassName(componentName) +
+                        jsmUriToClassName(jmsUri) +
+                        classNameSuffix;
+
+        return ClassName.get(basePackageName, simpleName);
+    }
+
+    private String jsmUriToClassName(final String jmsUri) {
```
mapingo: should be jmsUriToClassName instead of jsmUriToClassName

PR #567, opened 2018-04-10
Add catalog schema resolution to matchers

skiddykong:


pom.xml
```diff
@@ -58,6 +58,7 @@
         <utilities.version>1.12.0</utilities.version>
         <generator-maven-plugin.version>2.5.0-M2</generator-maven-plugin.version>
         <framework-api.version>3.0.0-M2</framework-api.version>
+        <json-schema-catalog.version>1.2.1</json-schema-catalog.version>
```
mapingo: there is a version set in the maven-parent-pom...we should update that

PR #579, opened 2018-05-01
Lazy load the mapping cache rather than use postConstruct

allanmckenzie:


core/src/test/java/uk/gov/justice/services/core/mapping/MappingCacheInitialiserTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class MappingCacheInitialiserTest {
+
+    @Mock
+    ActionNameToMediaTypesMappingObserver actionNameToMediaTypesMappingObserver;
```
mapingo: actionNameToMediaTypesMappingObserver could be private

core/src/test/java/uk/gov/justice/services/core/mapping/MappingCacheInitialiserTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Mock
+    BeanInstantiater beanInstantiater;
```
mapingo: beanInstantiater could be private

PR #581, opened 2018-05-02
Add more detailed schema validation exception description

skiddykong:
Also fixed bug in the schema validation logger as DI does not work there.

### DIFF
--- a/core/src/main/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandler.java
+++ b/core/src/main/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandler.java
@@ -18,13 +18,15 @@ public class LoggingValidationExceptionHandler implements EnvelopeValidationExce
 
     @Override
     public void handle(final EnvelopeValidationException ex) {
-
-        String validationTrace = "";
-
-        if(ex.getCause() instanceof ValidationException) {
-            validationTrace = defaultJsonValidationLoggerHelper.toValidationTrace((ValidationException) ex.getCause());
+        if(ex.getCause() instanceof ValidationException && logger.isWarnEnabled()) {
+            logger.warn(traceValidation(ex), ex);
+        } else if(logger.isWarnEnabled()) {
+            logger.warn("Message validation failed ", ex);
         }
+    }
 
-        logger.warn(format("Message validation failed %s", validationTrace), ex);
+    private String traceValidation(EnvelopeValidationException ex) {
+        return format("Message validation failed %s",
+                defaultJsonValidationLoggerHelper.toValidationTrace((ValidationException) ex.getCause()));
     }
 }
--- a/core/src/main/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandler.java
+++ b/core/src/main/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandler.java
@@ -1,18 +1,30 @@
 package uk.gov.justice.services.core.envelope;
 
-import javax.enterprise.inject.Alternative;
-import javax.inject.Inject;
+import static java.lang.String.format;
 
+import uk.gov.justice.services.core.json.DefaultJsonValidationLoggerHelper;
+
+import javax.enterprise.inject.Alternative;
+
+import org.everit.json.schema.ValidationException;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Alternative
 public class LoggingValidationExceptionHandler implements EnvelopeValidationExceptionHandler {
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/uk/gov/justice/services/core/json/FileBasedJsonSchemaValidator.java
```diff
@@ -28,6 +30,8 @@
     @Inject
     PayloadExtractor payloadExtractor;
 
+    DefaultJsonValidationLoggerHelper defaultJsonValidationLoggerHelper = new DefaultJsonValidationLoggerHelper();
```
mapingo: This can be injected as a JsonValidationLoggerHelper

core/src/main/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandler.java
```diff
@@ -1,18 +1,32 @@
...
+import org.everit.json.schema.ValidationException;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Alternative
 public class LoggingValidationExceptionHandler implements EnvelopeValidationExceptionHandler {
 
-    @Inject
-    private Logger logger;
+    private DefaultJsonValidationLoggerHelper defaultJsonValidationLoggerHelper = new DefaultJsonValidationLoggerHelper();
```
mapingo: This can be injected as a JsonValidationLoggerHelper

core/src/test/java/uk/gov/justice/services/core/json/FileBasedJsonSchemaValidatorTest.java
```diff
@@ -34,6 +36,10 @@
     @Mock
     private PayloadExtractor payloadExtractor;
 
+    @Mock
+    private DefaultJsonValidationLoggerHelper defaultJsonValidationLoggerHelper;
```
mapingo: This can be injected as a JsonValidationLoggerHelper

core/src/main/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandler.java
```diff
@@ -1,18 +1,32 @@
...
+    private Logger logger = LoggerFactory.getLogger(this.getClass());
```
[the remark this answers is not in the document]
skiddykong: As discussed we cannot inject here as producer uses reflection. Have tried alternative methods with the BeanManager, '@Vetoed' etc but CDI expects us to use beans.xml and we don't currently do that on project.

core/src/main/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandler.java
skiddykong: As discussed we cannot inject here as producer uses reflection. Have tried alternative methods with the BeanManager, '@Vetoed' etc but CDI expects us to use beans.xml and we don't currently do that on project.

core/src/test/java/uk/gov/justice/services/core/envelope/LoggingValidationExceptionHandlerTest.java
```diff
@@ -15,16 +20,28 @@
     @Mock
     private Logger logger;
 
+    @Mock
+    private DefaultJsonValidationLoggerHelper defaultJsonValidationLoggerHelper;
```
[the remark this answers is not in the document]
skiddykong: JsonValidationLoggerHelper does not have knowledge of ValidationExceptions so cannot work here.

core/src/test/java/uk/gov/justice/services/core/json/FileBasedJsonSchemaValidatorTest.java
skiddykong: JsonValidationLoggerHelper does not have knowledge of ValidationExceptions so cannot work here.

core/src/main/java/uk/gov/justice/services/core/json/FileBasedJsonSchemaValidator.java
skiddykong: JsonValidationLoggerHelper does not have knowledge of ValidationExceptions so cannot work here. I have now injected the DefaultJsonValidationLoggerHelper instead of creating.

skiddykong: Have tried to inject this but there is a brittle dependency on JsonSchemaValidatingMockTest.

PR #583, opened 2018-05-02
Add datasource producer.

MartinYSpasov:


event-sourcing/event-source/src/main/java/uk/gov/justice/services/eventsourcing/source/core/DataSourceProducer.java
```diff
@@ -0,0 +1,58 @@
...
+
+        if (eventSourceFor.isPresent()) {
+
+            final Location location = eventSourceFor.get().getLocation();
+
+            if (location.getDataSource().isPresent()) {
+                final Optional<String> dataSourceLocation = location.getDataSource();
+                return createDataSource(eventSourceName, dataSourceLocation);
+            } else {
+                return null;
```
allanmckenzie: I'd be tempted to throw an Exception here rather than returning null

event-sourcing/event-source/src/main/java/uk/gov/justice/services/eventsourcing/source/core/DataSourceProducerException.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.justice.services.eventsourcing.source.core;
+
+/**
+ * Exception thrown when there's error accessing database
+ */
+public class DataSourceProducerException extends RuntimeException {
+
+    private static final long serialVersionUID = 5934757152541630746L;
+
+    public DataSourceProducerException(String message) {
```
mapingo: message can be final

PR #611, opened 2018-05-18
Rename eventsources and subscritiondescriptor

MartinYSpasov:


MohamedFarouk-HMCTS:
Not required as it is now super seeded by another one

PR #625, opened 2018-06-11
fix for source update

MohamedFarouk-HMCTS:


event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/repository/streamstatus/StreamStatusJdbcRepository.java
```diff
@@ -126,6 +127,25 @@ public void update(final StreamStatus streamStatus) {
         }
     }
 
+    /**
+     * Returns a count of records for a given stream streamId.
+     *
+     * @param streamId streamId of the stream.
+     * @return a int.
+     */
+    public int countByStreamId(final UUID streamId) {
```
mapingo: Does this need to be production code as it is only used in the test...maybe this should be a test util

purple52:
Let's move the countByStreamId into the test itself before we merge into master.

event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/repository/streamstatus/StreamStatusJdbcRepository.java
MohamedFarouk-HMCTS: yes spoke to David and he understand the issue and said to change this as part of master branch release and leave it here for now

PR #627, opened 2018-06-12
fix for source update

MohamedFarouk-HMCTS:


event-buffer/event-buffer-core/src/main/java/uk/gov/justice/services/event/buffer/core/repository/streamstatus/StreamStatusJdbcRepository.java
```diff
@@ -126,6 +127,25 @@ public void update(final StreamStatus streamStatus) {
         }
     }
 
+    /**
+     * Returns a count of records for a given stream streamId.
+     *
+     * @param streamId streamId of the stream.
+     * @return a int.
+     */
+    public int countByStreamId(final UUID streamId) {
```
purple52: Since this is used only by the test, lets move this out into the test itself so it isn't available in the main code.

MohamedFarouk-HMCTS: sure

PR #643, opened 2018-07-10
Add ability to return Accepted and payload from Command API

mapingo:


generators/rest-adapter-generator/src/main/java/uk/gov/justice/services/adapters/rest/generator/ResponseStrategyFactory.java
```diff
@@ -0,0 +1,67 @@
...
+            return OK_STATUS_ENVELOPE_ENTITY_RESPONSE_STRATEGY;
+        }
+
+        return OK_STATUS_ENVELOPE_PAYLOAD_ENTITY_RESPONSE_STRATEGY;
+    }
+
+    private boolean containsOctetStreamResponse(final Action action) {
+        for (org.raml.model.Response response : action.getResponses().values()) {
+            if (response.getBody() != null) {
+                for (MimeType mimeType : response.getBody().values()) {
```
allanmckenzie: final?

PR #688, opened 2019-02-20
Add JMX service

MartinYSpasov:


jmx/src/main/java/uk/gov/justice/services/jmx/MBeanInstantiator.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.management.MBeanRegistrationException;
+import javax.management.MBeanServer;
+import javax.management.NotCompliantMBeanException;
+
+@Singleton
+@Startup
+public class MBeanInstantiator {
+
+    @Inject
+     MBeanRegistry mBeanRegistry;
```
maheshsubramanian: private

mapingo: We had a team decision, sometime ago, to make all injected classes package protected.  I believe we should continue to do so for consistency.

PR #723, opened 2019-05-03
move yaml parser to utilities

MohamedFarouk-HMCTS:


event-subscription/pom.xml
```diff
@@ -37,7 +37,11 @@
             <groupId>com.github.everit-org.json-schema</groupId>
             <artifactId>org.everit.json.schema</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>uk.gov.justice.utils</groupId>
+            <artifactId>utilities-core</artifactId>
+            <version>${utilities.version}</version>
```
mapingo: is the version needed?

event-subscription/src/main/java/uk/gov/justice/subscription/SubscriptionHelper.java
```diff
@@ -25,7 +25,7 @@ private Integer getPrioritisationNumber(final String prioritisation) {
             }
             return HIGHEST_PRIORITY;
         } catch (final NumberFormatException nfe) {
-            throw new YamlParserException(format("Incorrect prioritisation number: %s in subscription-descriptor.yaml", prioritisation));
+            throw new YamlParserException(format("Incorrect prioritisation number: %s in subscription-descriptor.yaml", prioritisation),nfe);
```
mapingo: minor...but there could be a space after the comma

pom.xml
```diff
@@ -46,7 +46,7 @@
         <embedded-artemis.version>1.2.0</embedded-artemis.version>
         <jboss-ejb3-ext-api.version>2.2.0.Final</jboss-ejb3-ext-api.version>
         <test-utils.version>1.22.0-M1</test-utils.version>
-        <utilities.version>1.16.4</utilities.version>
+        <utilities.version>1.17.0-M3</utilities.version>
```
mapingo: maybe we should have released utilities as a full release rather than Milestone release...we should sort that out later

event-subscription/pom.xml
MohamedFarouk-HMCTS: removed

pom.xml
MohamedFarouk-HMCTS: fixed

event-subscription/src/main/java/uk/gov/justice/subscription/SubscriptionHelper.java
MohamedFarouk-HMCTS: done

PR #743, opened 2019-06-24
Fix problem of shuttering being called twice on catchup

allanmckenzie:


framework-utilities/src/test/java/uk/gov/justice/services/framework/utilities/cdi/CdiInstanceResolverTest.java
```diff
@@ -42,6 +42,6 @@ public void shouldResolvedTheInstanceOfTheBean() throws Exception {
                 beanClass,
                 context)).thenReturn(systemCommandScanner);
 
-        assertThat(cdiInstanceResolver.getInstanceOf(beanClass, beanManager), is(systemCommandScanner));
+        assertThat(cdiInstanceResolver.getInstanceOf(beanClass, beanManager), CoreMatchers.is(systemCommandScanner));
```
mapingo: CoreMatchers.is can be static import

pom.xml
```diff
@@ -49,7 +50,7 @@
         <test-utils.version>1.23.0</test-utils.version>
         <utilities.version>1.19.0</utilities.version>
         <json-schema-catalog.version>1.7.0</json-schema-catalog.version>
-        <framework-api.version>4.0.0-M25</framework-api.version>
+        <framework-api.version>4.0.0</framework-api.version>
```
mapingo: This will need to be the M25 release for now

PR #777, opened 2019-10-01
Add observer for the lifecycle of a SystemCommand

allanmckenzie:


jmx/jmx-command-handling/pom.xml
```diff
@@ -40,6 +45,11 @@
             <artifactId>persistence-jdbc</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+            <scope>test</scope>
+        </dependency>
```
mapingo: minor....this could be moved below the Test Dependencies comment

